PR #1, opened 2021-08-18
load proposals for realm and display as list

mschneider:


stores/useWalletStore.tsx
```diff
@@ -29,54 +42,227 @@ interface WalletStore extends State {
...
+          const voteRecords = await getGovernanceAccounts<VoteRecord>(
+            programId,
+            endpoint,
+            VoteRecord,
+            getAccountTypes(VoteRecord),
+            [pubkeyFilter(1, new PublicKey(p))]
+          )
+
+          const precision = Math.pow(10, 6)
+          const yes =
```
SebastianBor, line 233: There is no need to manually calculate yes/no votes. You have `yesVotesCount` and `noVotesCount` on `Proposal`

mschneider: good catch! removed it

PR #2, opened 2021-08-19
notifier for governance proposals just open for voting

microwavedcola1:
Here is an attempt at the discord notifier, the simple script scans for mgno proposals opened in last 5 mins, and notifies to a WEBHOOK_URL of choice

### DIFF
--- a/scripts/governance-notifier.ts
+++ b/scripts/governance-notifier.ts
@@ -74,16 +74,16 @@ async function runNotifier() {
       // voting is closed
       proposal.info.votingCompletedAt
     ) {
-      continue
       countSkipped++
+      continue
     }
 
     if (
       // not yet signed i.e. only in draft
       !proposal.info.signingOffAt
     ) {
-      continue
       countSkipped++
+      continue
     }
 
     if (
--- a/scripts/governance-notifier.ts
+++ b/scripts/governance-notifier.ts
@@ -59,31 +59,27 @@ async function runNotifier() {
   )
 
   const realmProposals = Object.fromEntries(
-    Object.entries(proposals)
-      .filter(
-        ([_k, v]) =>
-          Object.keys(realmGovernances).includes(
-            v.info.governance.toBase58()
-          ) && v.info.votingAtSlot
-      )
-      .sort(
-        (a, b) => b[1].info.votingAt.toNumber() - a[1].info.votingAt.toNumber()
-      )
+    Object.entries(proposals).filter(
+      ([_k, v]) =>
+        Object.keys(realmGovernances).includes(v.info.governance.toBase58()) &&
+        v.info.votingAtSlot
+    )
   )
 
   for (const k in realmProposals) {
     const proposal = realmProposals[k]
+
     if (
       // voting is closed
       proposal.info.votingCompletedAt
     )
-      return
+      continue
 
     if (
       // not yet signed i.e. only in draft
-      proposal.info.signingOffAt === null
+      !proposal.info.signingOffAt
     )
-      return
+      continue
 
     const fiveMinutesSeconds = 5 * 60
     if (
--- a/scripts/governance-notifier.ts
+++ b/scripts/governance-notifier.ts
@@ -2,11 +2,9 @@ import { PublicKey } from '@solana/web3.js'
 import axios from 'axios'
 import { RealmInfo } from '../@types/types'
 import { getAccountTypes, Governance, Proposal } from '../models/accounts'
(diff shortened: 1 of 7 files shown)


scripts/governance-notifier.ts
```diff
@@ -0,0 +1,101 @@
...
+  const realmProposals = Object.fromEntries(
+    Object.entries(proposals)
+      .filter(
+        ([_k, v]) =>
+          Object.keys(realmGovernances).includes(
+            v.info.governance.toBase58()
+          ) && v.info.votingAtSlot
+      )
+      .sort(
+        (a, b) => b[1].info.votingAt.toNumber() - a[1].info.votingAt.toNumber()
```
mschneider, line 70: you might not need this sort

scripts/governance-notifier.ts
```diff
@@ -0,0 +1,101 @@
...
+    )
+  )
+
+  const realmProposals = Object.fromEntries(
+    Object.entries(proposals)
+      .filter(
+        ([_k, v]) =>
+          Object.keys(realmGovernances).includes(
+            v.info.governance.toBase58()
+          ) && v.info.votingAtSlot
```
mschneider, line 67: the last filter condition is also not needed anymore

scripts/governance-notifier.ts
```diff
@@ -0,0 +1,101 @@
+import { PublicKey } from '@solana/web3.js'
+import axios from 'axios'
+import { RealmInfo } from '../@types/types'
+import { getAccountTypes, Governance, Proposal } from '../models/accounts'
+import { getGovernanceAccounts, pubkeyFilter } from '../models/api'
+import { ParsedAccount } from '../models/serialisation'
+import { ENDPOINTS } from '../stores/useWalletStore'
+
+global.fetch = require('node-fetch')
```
mschneider, line 9: image

scripts/governance-notifier.ts
microwavedcola1: actually if I remove the `&& v.info.votingAtSlot`, then I also get the draft proposals

scripts/governance-notifier.ts
```diff
@@ -0,0 +1,101 @@
...
+      fiveMinutesSeconds
+    ) {
+      const msg = `${proposal.info.name} proposal just opened for voting`
+      console.log(msg)
+      axios.post(process.env.WEBHOOK_URL, { msg })
+    }
+  }
+}
+
+runNotifier()
```
mschneider, line 101: run with setInterval

scripts/governance-notifier.ts
microwavedcola1: fixed

scripts/governance-notifier.ts
microwavedcola1: I would like to re-use the code from the governance-ui, hence this ugly statement 😅

mschneider (approved):

scripts/governance-notifier.ts
```diff
@@ -0,0 +1,107 @@
...
+      // not yet signed i.e. only in draft
+      !proposal.info.signingOffAt
+    ) {
+      countSkipped++
+      continue
+    }
+
+    if (
+      // proposal opened in last 5 mins
+      nowInSeconds - proposal.info.signingOffAt.toNumber() <=
```
SebastianBor, line 91: Voting starts when `votingAt` is set and not `signingOffAt`. `signingOffAt` is set when signatories start signing off a proposal. If you have just one signatory it'is going to be the same timestamp as `votingAt` but it doesn't have to be in general case

scripts/governance-notifier.ts
```diff
@@ -0,0 +1,107 @@
...
+    )
+  )
+
+  console.log(`- scanning all proposals`)
+  let countJustOpened = 0
+  let countSkipped = 0
+  for (const k in realmProposals) {
+    const proposal = realmProposals[k]
+
+    if (
```
SebastianBor, line 73: You can filter out all proposals in pre voting state using `proposal.isPreVotingState()`

scripts/governance-notifier.ts
```diff
@@ -0,0 +1,107 @@
...
+      !proposal.info.signingOffAt
+    ) {
+      countSkipped++
+      continue
+    }
+
+    if (
+      // proposal opened in last 5 mins
+      nowInSeconds - proposal.info.signingOffAt.toNumber() <=
+      fiveMinutesSeconds
```
SebastianBor, line 92: I think it would be good to add some tolerance to accommodate for potential time drift and difference in clocks

PR #3, opened 2021-08-24
fetch proposal data

SebastianBor:


models/api.ts
```diff
@@ -179,3 +183,14 @@ async function getGovernanceAccountsImpl<TAccount extends GovernanceAccount>(
...
+export async function getGovernanceAccount<TAccount extends GovernanceAccount>(
+  connection: Connection,
+  accountPubKey: PublicKey,
+  accountClass: GovernanceAccountClass
+) {
+  const accountInfo = await connection.getAccountInfo(accountPubKey)
+  const account = BorshAccountParser(accountClass)(accountPubKey, accountInfo)
+
+  return account as ParsedAccount<TAccount>
+}
```
mschneider, line 196: hah! that's a nice one, i'll use it as well :)

SebastianBor: :)

PR #13, opened 2021-08-28
comment before confirming vote

saml33:
I haven't tested this as I couldn't figure out how to get off mainnet 🙃


components/DiscussionForm.tsx
```diff
@@ -1,16 +1,51 @@
...
+      connection.endpoint
+    )
+
+    const msg = new ChatMessageBody({
+      type: ChatMessageBodyType.Text,
+      value: comment,
+    })
+
+    try {
+      await postChatMessage(rpcContext, proposal, ownTokenRecord.pubkey, msg)
```
SebastianBor, line 39:
```suggestion
      await postChatMessage(rpcContext, proposal, ownTokenRecord.pubkey, msg)
      setComment('')
```

SebastianBor (commented):
The vote confirmation popup seems to me to be too small and cramped for such a significant action. Would it make sense to make it bigger?
Or maybe we can avoid it altogether and add the vote comment directly on the voting panel?

components/VoteCommentModal.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    try {
+      await castVote(
+        rpcContext,
+        realm.pubkey,
+        proposal,
+        ownTokenRecord.pubkey,
+        vote
+      )
+      if (comment) {
```
SebastianBor, line 58: It would be better UX to combine vote and chat instruction into a single transaction to avoid two wallet confirmations

saml33:
> The vote confirmation popup seems to me to be too small and cramped for such a significant action. Would it make sense to make it bigger?
> Or maybe we can avoid it altogether and add the vote comment directly on the voting panel?

Yeah, it still needs a little work. Assume there is a max-length for a comment? The input should probably be a textarea too. I think it's cleaner to just have the vote buttons in the vote panel as there is already the discussion form above. It does feel like an extra step though.

SebastianBor:
> > The vote confirmation popup seems to me to be too small and cramped for such a significant action. Would it make sense to make it bigger?
> > Or maybe we can avoid it altogether and add the vote comment directly on the voting panel?
> 
> Yeah, it still needs a little work. Assume there is a max-length for a comment? The input should probably be a textarea too. I think it's cleaner to just have the vote buttons in the vote panel as there is already the discussion form above. It does feel like an extra step though.

I'm wondering whether the vote panel could be on the side below the deposit panel to have the actionable UI in one place. Then comment on the vote panel wouldn't look that bad. And maybe it could be blow the voting buttons?

PR #14, opened 2021-08-28
correctly calculate vote progress

mschneider:


hooks/useProposalVotes.tsx
```diff
@@ -0,0 +1,37 @@
+import { Proposal } from '../models/accounts'
+import { calculatePct, fmtTokenAmount } from '../utils/formatting'
+import useRealm from './useRealm'
+
+export default function useProposalVotes(proposal?: Proposal) {
+  const { mint, governances } = useRealm()
```
SebastianBor, line 6: mint should be taken from the proposal. It's irrelevant for MNGO and we can ignore for now because there is no council but in general case either council or community can vote on a proposal

SebastianBor (approved):

PR #16, opened 2021-08-28
add proposal vote status to list view

saml33:


SebastianBor:
Looks nice, especially on the Proposal view.  Some small display issues 

1) Percentages should be formatted to 0.01 precision and if value is below that I would show '<0.1%' or something similar 
2) If the No vote percentage is too small the bar end is not rounded. I think we should always show the rounded end if there is any value

image

saml33:
> Looks nice, especially on the Proposal view. Some small display issues
> 
> 1. Percentages should be formatted to 0.01 precision and if value is below that I would show '<0.1%' or something similar
> 2. If the No vote percentage is too small the bar end is not rounded. I think we should always show the rounded end if there is any value
> 
> image

Cool, thanks man. I hadn't tested this on mainnet yet so didn't see the issues you have pointed out. All the percentages on dev were 100 or 0

SebastianBor (approved):
Looks good!

PR #34, opened 2021-09-27
realms + realm switch

abrzezinski94:


pages/index.tsx
```diff
@@ -5,7 +5,7 @@ const Index = () => {
   const router = useRouter()
 
   useEffect(() => {
-    router.push('/dao/MNGO')
+    router.push('/realms')
```
mschneider, line 8: should be configurable via ENV variable for deployment

models/registry/api.ts
```diff
@@ -33,29 +33,33 @@ const MAINNET_REALMS: RealmInfo[] = [
...
-    ogImage: 'https://token.mango.markets/preview.jpg',
+    ogImage: 'https://www.phantom.app/img/logo.png',
   },
 
   {
     symbol: 'SOCEAN',
     programId: new PublicKey('5hAykmD4YGcQ7Am3N7nC9kyELq6CThAkU82nhNKDJiCy'),
     realmId: new PublicKey('759qyfKDMMuo9v36tW7fbGanL63mZFPNbhU7zjPrkuGK'),
     website: 'https://www.socean.fi',
+    ogImage: 'https://www.phantom.app/img/logo.png',
```
mschneider, line 44: why should this be changed?

mschneider: this is the official mango logo: https://mango.markets/img/logo_mango.svg

mschneider:
Hello Adrian, thank you for the PR. I asked @stevesarmiento to make a nice default logo / placeholder for a governance that doesn't define it's own yet. He should send you a link later today.

stevesarmiento:
@abrzezinski94 Hey Adrian, lets try this in there :)
govern

stevesarmiento:
Alternativel
govern2
y

SebastianBor:
> @abrzezinski94 Hey Adrian, lets try this in there :)
> govern

This feels too much like a government logo to me, while realms are self organised organisations. Example of realms:
- DAO over on-chain protocol - MNGO
- Governance over stake pool - SOCEAN
- On chain hedge fund - SCTF1, Yield Farming
- On chain multiverse game
- Group of people trading NFTs together

mschneider:
what do you think about the heroicons user group one?

image

SebastianBor:
> what do you think about the heroicons user group one?
> 
> image

Yes, I was thinking along those lines, like the old SharePoint server logo

image

pages/index.tsx
abrzezinski94: If i understand right we want to have realms page or eg /dao/mngo page as home page if i chose to have dao/mngo as my home page with the build, should users have access to realms page if they typed the url or should it be completely hidden from them ?

SebastianBor: Yeah, I think they should have access to the full ecosystem. It's very likely users would participate in many realms and this is why we are building this one stop shop for them.

mschneider:
Don’t fully agree with this.

1) There should definitely be a version that is hosted independently and
offers all governances. Similar to Snapshot orgs offering. I see that a lot
of users would use this in the beginning.

2) Still individual teams should be able to run their own version including
customized UI. Reason here is that most protocols should actually heavily
customize governance and spend development time on it. The realms overview
page really confused Mango users and are immediately caused issues inside
our community, people didn’t feel at home when they saw a new realm popup.
For a general service it needs to be a central element in the navigation
for a whitelabel version it shouldn’t exist.

If customization is not easy, people will fork or rewrite. We can’t prevent
it, but we should aim for extensibility and supporting the communities
features rather then enforcing equality, so the incentives are minimized.

SebastianBor:
> Don’t fully agree with this. 1) There should definitely be a version that is hosted independently and offers all governances. Similar to Snapshot orgs offering. I see that a lot of users would use this in the beginning. 2) Still individual teams should be able to run their own version including customized UI. Reason here is that most protocols should actually heavily customize governance and spend development time on it. The realms overview page really confused Mango users and are immediately caused issues inside our community, people didn’t feel at home when they saw a new realm popup. For a general service it needs to be a central element in the navigation for a whitelabel version it shouldn’t exist. If customization is not easy, people will fork or rewrite. We can’t prevent it, but we should aim for extensibility and supporting the communities features rather then enforcing equality, so the incentives are minimized.
> […](#)

Yeah, I see were you are coming from with this. So, let's do this:

1) If there is no env variable then we show the realm header with back navigation to all realms
2) If the env variable is set then we hide the header and navigation and use the realm logo as the page logo (no Solana logo)

How does it sound?

mschneider:
tried merging your newest change with @saml33 PR https://github.com/blockworks-foundation/governance-ui/pull/37 but couldn't finish it. could you have a look @abrzezinski94 ?

i'll update our deployment target for dao.mango.markets to set the env varibale and spin up a new deployment under a different domain so we can test both.

mschneider:
sad-einstein: MNGO
wizardly-kowalevski-4abe4b: all realms

abrzezinski94:
@mschneider I've added the changes from @saml33 #37 pr

PR #38, opened 2021-10-04
Deposit custom amount

saml33:
Something isn't quite right with the deposit amount. Probably using BN incorrectly...


mschneider:
had a look at this just now, looks like the goverance program api wrapper does not expose a way to set the amount, also checked the governance program and looks like the respective instruction does just always transfer everything. can't merge this until we update the governance program with that functionality

PR #62, opened 2021-11-05
Create realm support

jpbogle:
hi @mschneider, @SebastianBor 

I was playing around with the UI and figured I could help make this PR that is a bit more silo-ed 

I was seeing that the withCreateRealm wasn't hooked up to an action at all so I decided it would be useful to hook that up for if someone wants to create new realms through this UI since creating the raw instruction is obviously hard manually.

image
image

I wanted to open a draft to hear thoughts.

I also looked at the other UI example https://github.com/solana-labs/oyster/blob/main/packages/governance/src/models/withCreateRealm.ts#L36 and I see the use of programVersion. Not sure we would want to include that in this version of the UI or if that version is too old now that all new governance program instances wouldnt use it

I needed this functionality to use this UI to create a new realms. It doesnt need to be hooked up to a big button on the realms page if not desired

### DIFF
--- a/models/serialisation.ts
+++ b/models/serialisation.ts
@@ -89,6 +89,7 @@ export const GOVERNANCE_SCHEMA = new Map<any, any>([
         ['useCouncilMint', 'u8'],
         ['minCommunityTokensToCreateGovernance', 'u64'],
         ['communityMintMaxVoteWeightSource', MintMaxVoteWeightSource],
+        ['useCommunityVoterWeightAddin', 'u8'],
       ],
     },
   ],
--- a/models/serialisation.ts
+++ b/models/serialisation.ts
@@ -39,10 +39,12 @@ import {
   SignatoryRecord,
   TokenOwnerRecord,
   VoteRecord,
+  RealmConfigAccount,
   VoteThresholdPercentage,
   VoteWeight,
 } from './accounts'
 import { serialize } from 'borsh'
+import { ProgramVersion } from './registry/api'
 
 // Temp. workaround to support u16.
 ;(BinaryReader.prototype as any).readU16 = function () {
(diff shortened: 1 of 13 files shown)


SebastianBor:
> I needed this functionality to use this UI to create a new realms. It doesnt need to be hooked up to a big button on the realms page if not desired

Thank you for the contribution. It's appreciated and much needed. 
It's busy time right now with the conference in Lisbon so I'll only reply very quickly and then get back to it with more details after the conference next week.

1) The version parameter is needed, in fact all the instances on mainnet and devnet I'm aware of are on v1
     In v2 there is slight change in the api, but I haven't ported the changes from the oyster repo yet

2) The UI currently maintains hardcoded registry of the governance program clones and their realms and metadata. This should be moved on-chain to fully support the workflow to create realms through the UI.
If we are not going to support it then the UI at minimum should provide explanation saying it only creates the realm and the extra config steps as temp. measures are required to see the realm in the UI

jpbogle:
>It's busy time right now with the conference in Lisbon so I'll only reply very quickly and then get back to it with more details after the conference next week.

yep of course. 

1. got it
2. Ya was thinking we need a message saying realm was created and has to be added to the list. Could look into looking up all realms for each program instance from chain dynamically in this PR or a follow, but not sure where the list of programs should come from still if you just created a new one

pages/realms/new.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const connection = useWalletStore((s) => s.connection)
+
+  const [form, setForm] = useState({
+    governanceProgramId: '',
+    name: '',
+    communityTokenMint: '',
+    councilMint: '',
+    communityMintMaxVoteWeightSource: 0.5,
+    minCommunityTokensToCreateGovernance: 0.001,
+  })
```
jpbogle, line 79: appropriate defaults should be selected here

pages/realms/new.tsx
```diff
@@ -0,0 +1,274 @@
...
+    .string()
+    .required('Community token mint is required')
+    .test(
+      'is-public-key',
+      'Community token mint id is not a valid public key',
+      publicKeyValidationTest
+    ),
+})
+
+const defaultGovernanceCtx: InstructionsContext = {
```
abrzezinski94, line 53: probably not needed here

jpbogle: thanks ya this was unused

models/serialisation.ts
```diff
@@ -80,411 +82,446 @@ export const serializeInstructionToBase64 = (
...
-    },
-  ],
+export function getGovernanceSchema(programVersion: ProgramVersion) {
+  switch (programVersion) {
+    case ProgramVersion.V1:
+      return GOVERNANCE_SCHEMA
+    default:
+      return GOVERNANCE_SCHEMA_V2
+  }
+}
```
jpbogle, line 95: had to port this over from realm-explorer to have it work properly for V2 governance programs

pages/realms/new.tsx
SebastianBor: Please use the following defaults:
`communityMintMaxVoteWeightSource`: 1 - by default we want to use full mint supply as max vote weight
`minCommunityTokensToCreateGovernance`: let's use 1% of the community token supply, and if the supply is 0 let's make it arbitrary `1000000` value. 
`governnaceProgramId`: `GovER5Lthms3bLBqWub97yVrMmEogzX7xNjdXpPPCVZw` the default instance

models/serialisation.ts
SebastianBor: :+1:

SebastianBor:
> > It's busy time right now with the conference in Lisbon so I'll only reply very quickly and then get back to it with more details after the conference next week.
> 
> yep of course.
> 
> 1. got it  - and ya haha i guess the instance i just deployed is v2 so i didnt know 😅
> 2. Ya was thinking we need a message saying realm was created and has to be added to the list. Could look into looking up all realms for each program instance from chain dynamically in this PR or a follow, but not sure where the list of programs should come from still if you just created a new one

The list of governance clones should be coming from on-chain governance programs registry. Since the program doesn't exist yet we can just add the comment in this PR explain the situation.

let's reorder the inputs in the form:

`PorgramId/Version` - should be the last 2 inputs
`Min community tokens to create governance` - below community mint

pages/realms/new.tsx
jpbogle: ya makes sense - for the 1% of community mint, should we use 1% of council mint if specified?

1% council mint || 1% community mint || 1000000

PR #76, opened 2021-11-17
Use connection config from the wallet store

jpbogle:
Connection is currently set in useWalleStore, but it used separately for fetchRealms

This PR fixes it so the connection in useWalletStore is also used in fetchRealms and it gets set in hydrateStore so the same connection can be used elsewhere and set via the cluster= param.

also added localnet support

### DIFF
--- a/hooks/useRealm.tsx
+++ b/hooks/useRealm.tsx
@@ -8,8 +8,8 @@ import useWalletStore from '../stores/useWalletStore'
 
 export default function useRealm() {
   const router = useRouter()
-  const { symbol, cluster } = router.query
-  const apiEndpoint = cluster ? (cluster as EndpointTypes) : 'mainnet'
+  const { symbol } = router.query
+  const connection = useWalletStore((s) => s.connection)
   const connected = useWalletStore((s) => s.connected)
   const wallet = useWalletStore((s) => s.current)
   const tokenAccounts = useWalletStore((s) => s.tokenAccounts)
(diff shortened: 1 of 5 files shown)


hooks/useRealm.tsx
```diff
@@ -25,9 +25,11 @@ export default function useRealm() {
...
     councilTokenOwnerRecords,
   } = useWalletStore((s) => s.selectedRealm)
-  const realmInfo = useMemo(() => getRealmInfo(symbol as string, apiEndpoint), [
-    symbol,
-  ])
+
+  const realmInfo = useMemo(
+    () => getRealmInfo(symbol as string, connection.endpoint as EndpointTypes),
+    [symbol]
+  )
```
jpbogle, line 32: This is the change to use the connection from useWalletStore for getting the realms and realmInfo, so that it all uses the same connection (set from query param)

stores/useWalletStore.tsx
```diff
@@ -192,7 +198,7 @@ const INITIAL_PROPOSAL_STATE = {
 
 const useWalletStore = create<WalletStore>((set, get) => ({
   connected: false,
-  connection: INITIAL_CONNECTION_STATE,
+  connection: getConnectionConfig('mainnet'),
```
jpbogle, line 201: making this as default means the UI will start to render with mainnet.

If something else is set via query param, it will immediately re-render. A fix for this is to make connection optionally undefined. This can be done in a follow-up PR since its a messier change and we may want to consider another approach like making connection its own context

SebastianBor: The extra re-render for non mainnet is ok for now. Good triage!

hooks/useRealm.tsx
SebastianBor: :+1:

stores/useWalletStore.tsx
SebastianBor: It's ok to re-render for non mainnet. Good Triage! :+1:

SebastianBor:
Very good and much needed change. Thank you

PR #88, opened 2021-11-21
Only set connection if it changes https://github.com/blockworks-found…

jpbogle:
…ation/governance-ui/issues/83

fixes https://github.com/blockworks-foundation/governance-ui/issues/83

@SebastianBor @abrzezinski94 @dahifi 

since useHydrateStore depended on [cluster] I guess the next router considers cluster changing everytime a page changes so it was recreating a new connection which was disconnecting the wallet. To fix this i put in a check to only re-create connection if it has actually changed. We could put this check inside useHydrateStore but i put it directly in useWalletStore so its more core to the wallet store

### DIFF
--- a/hooks/useHydrateStore.tsx
+++ b/hooks/useHydrateStore.tsx
@@ -17,8 +17,8 @@ export default function useHydrateStore() {
   } = useWalletStore((s) => s.actions)
   useEffect(() => {
     const fetch = async () => {
-      const realmInfo = getRealmInfo(symbol as string, apiEndpoint)
       setConnectionConfig(apiEndpoint)
+      const realmInfo = getRealmInfo(symbol as string, apiEndpoint)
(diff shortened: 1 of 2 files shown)


hooks/useHydrateStore.tsx
```diff
@@ -17,8 +17,8 @@ export default function useHydrateStore() {
   } = useWalletStore((s) => s.actions)
   useEffect(() => {
     const fetch = async () => {
-      const realmInfo = getRealmInfo(symbol as string, apiEndpoint)
       setConnectionConfig(apiEndpoint)
+      const realmInfo = getRealmInfo(symbol as string, apiEndpoint)
```
jpbogle, line 21: just moved this since it related to the code below

SebastianBor: :+1:

PR #95, opened 2021-11-22
Feature/90 issue fix

abrzezinski94:


jpbogle (approved):

pages/dao/[symbol]/proposal/components/instructions/ProgramUpgrade.tsx
```diff
@@ -102,7 +102,14 @@ const ProgramUpgrade = ({
...
   }, [form])
-
+  useEffect(() => {
+    if (governedProgramAccounts.length === 1) {
+      handleSetForm({
+        value: governedProgramAccounts[0],
+        propertyName: 'governedAccount',
+      })
+    }
+  }, [governedProgramAccounts.length])
```
Comment (jpbogle) [inline pages/dao/[symbol]/proposal/components/instructions/ProgramUpgrade.tsx:112]: so this is just to auto fill in that field if it knows there is only one right, so there is no need for fallback the user can just select one otherwise?

jpbogle (approved):

pages/dao/[symbol]/proposal/components/instructions/ProgramUpgrade.tsx
Comment (abrzezinski94) [reply pages/dao/[symbol]/proposal/components/instructions/ProgramUpgrade.tsx:112]: yes

Comment (abrzezinski94) [reply pages/dao/[symbol]/proposal/components/instructions/ProgramUpgrade.tsx:112]: oh i will change that i have idea for better solution

PR #97, opened 2021-11-24
Create realm

jpbogle:
image

image

image


actions/registerRealm.ts
```diff
@@ -0,0 +1,44 @@
...
+    undefined
+  )
+
+  const transaction = new Transaction()
+  transaction.add(...instructions)
+
+  return sendTransaction({ transaction, wallet, connection })
+}
+
+export async function getRealmIdFromTransaction(
```
SebastianBor, line 38: There is no need for this function. `registerRealm` can return the new realm address the same way as `createProposal` does

pages/realms/new.tsx
```diff
@@ -0,0 +1,391 @@
...
+      const mint = await tryGetMint(connection.current, mintPublicKey)
+      if (mint) {
+        const supply = mint.account.supply
+        const decimals = mint.account.decimals
+        if (supply.gt(new BN(0))) {
+          handleSetForm({
+            minCommunityTokensToCreateGovernance: BN.max(
+              new BN(1),
+              // divide by 10^(decimal + 2) to get 1% of supply
+              supply.div(new BN(10).pow(new BN(decimals + 2)))
```
SebastianBor, line 155: It's better to use `BigNumber` and `shiftedBy()` instead of div/mul operations to not loose precision. See `units.ts`  for examples

SebastianBor (approved):
I'm merging this PR because we need the api changes on master, but please address my comments retrospectively in a follow up PR.

SebastianBor:
> image

This confirmation screen is not displayed after a realm is created.

pages/realms/new.tsx
jpbogle: hmm looked for that ill look again

actions/registerRealm.ts
jpbogle: right - i thought it was a bit weird to not return tx-id for clicking on the transaction. it could return both. ill change that but ultimately it will just redirect back to the main page and reload the full list

pages/realms/new.tsx
jpbogle: oh its cuz units.ts uses BigNumber vs BN haha

PR #99, opened 2021-11-25
Address comments about realmAddress returned from transaction and spl…

jpbogle:
…ash screen not showing

1. remove extra RPC call to get the realmId by just returning and using the derived one from the transaction call
2. use units.ts for messing around with BN and bignumber

### DIFF
--- a/pages/realms/new.tsx
+++ b/pages/realms/new.tsx
@@ -8,7 +8,7 @@ import React, { useState } from 'react'
 import Button from '@components/Button'
 import { RpcContext } from '@models/core/api'
 import { MintMaxVoteWeightSource } from 'models/accounts'
-import { getRealmIdFromTransaction, registerRealm } from 'actions/registerRealm'
+import { registerRealm } from 'actions/registerRealm'
 import useWalletStore from 'stores/useWalletStore'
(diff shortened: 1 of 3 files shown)


pages/realms/new.tsx
```diff
@@ -146,13 +151,16 @@ const New = () => {
...
             minCommunityTokensToCreateGovernance: BN.max(
               new BN(1),
-              // divide by 10^(decimal + 2) to get 1% of supply
-              supply.div(new BN(10).pow(new BN(decimals + 2)))
+              // divide by 100 for a percentage
+              new BN(
+                getMintDecimalAmount(mint.account, supply)
+                  .dividedBy(100)
+                  .toString()
+              )
```
jpbogle, line 163: didnt shiftedBy before since it wasnt on BN but was on BigNumber haha

So I used units.ts for getting the BigNumber decimal amount and dividing but then I have to convert it back to BN unless we want to use BigNumber everywhere. Then there is the serum/anchor "BN" and the regular "BN" as well haha

jpbogle: sticking to regular BN for now since thats what most of the API layer uses rn

PR #102, opened 2021-11-26
fix: start notifier immediately (don't wait 5m initially)

johnrees:
Currently, after running `yarn notifier`, nothing will happen for 5 minutes because it waits for the first `setInterval` to fire.

I don't know if this is desired behaviour? If it isn't then this will call the script immediately without the initial 5 minute delay.

### DIFF
--- a/scripts/governance-notifier.ts
+++ b/scripts/governance-notifier.ts
@@ -107,4 +107,7 @@ async function runNotifier() {
   )
 }
 
+// start notifier immediately
+errorWrapper()
+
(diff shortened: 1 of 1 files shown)


johnrees:
Or maybe the delay is desired because of the 30s (toleranceSeconds) window?

I guess it is making the assumption that the script crashed and restarted within 30s, and has then needed to wait 5 minutes before checking again?

microwavedcola1 (approved):

Issue #103, opened 2021-11-26
Notifier not fetching all proposals due to RPC rate limits


johnrees:
The governance notifier is currently using the [public RPC endpoint](https://github.com/blockworks-foundation/governance-ui/blob/4df37f710ae3715648dc920f026efda961c95b5d/scripts/governance-notifier.ts#L23) and is being rate limited on the [getProgramAccounts](https://github.com/blockworks-foundation/governance-ui/blob/4df37f710ae3715648dc920f026efda961c95b5d/scripts/api.ts#L100) instruction.

There is no retry logic when this happens the data it uses is incomplete. I think it either needs to implement retry or throttling behaviour, or use a different endpoint.
 
## public rpc - api.mainnet-beta.solana.com
```
Unexpected response {"jsonrpc":"2.0","error":{"code":429,"message":"Too requests for a specific RPC call, contact your app developer or support@rpcpool.com."},"id":1}

-- countJustOpenedForVoting: 0, countVotingNotStartedYet: 19, countClosed: 43
```

## genesysgo rpc
`-- countJustOpenedForVoting: 0, countVotingNotStartedYet: 20, countClosed: 45`

## [theindex.io](https://theindex.io)
`-- countJustOpenedForVoting: 0, countVotingNotStartedYet: 20, countClosed: 45`

johnrees:
NB: there is no noticeable speed difference between the requests made to theindex.io and other RPC services in this instance

johnrees:
Ah, I just found something in the code that might fix it. BRB!

mschneider:
yeah, we never bother in working around rate limits and always just set a custom RPC via env. feel free to add another default that works better.

johnrees:
> yeah, we never bother in working around rate limits and always just set a custom RPC via env. feel free to add another default that works better.

before seeing this comment I posted PR that fixes rate limits here #104

or an faster alternative WIP PR if you're planning to use private RPC infra here #105

PR #105, opened 2021-11-26
remove rate limits from api

johnrees:
related #104 

An alternative to #103 which removes rate limits from the API and runs requests in parallel.

This will fail using public RPC infra due to rate-limiting, but runs in about 2-3s, rather than #103's ~15s, when using my own RPC provided via an env var i.e.

```
RPC_URL=$RPC yarn notifier
```

### DIFF
--- a/scripts/api.ts
+++ b/scripts/api.ts
@@ -68,13 +68,8 @@ export async function getGovernanceAccounts<TAccount extends GovernanceAccount>(
     )
   }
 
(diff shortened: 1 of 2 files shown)


microwavedcola1 (approved):

johnrees:
closed for now (using #104 instead) as this would break public RPC compatibility due to rate-limiting

PR #109, opened 2021-11-29
Groundwork for gov-registry - setting up fetching realms to be async …

jpbogle:
…and support lookup by realmId

@SebastianBor @mschneider @abrzezinski94 

some initial work for the stuff im working on in https://github.com/blockworks-foundation/governance-ui/tree/gov-registry

specifically this is a bit of refactoring to pass in a connectionContext to fetching realms, and making it async so that the fetching can be plopped in

and also supports `.../dao/{realmId}` in addition to `.../dao/{symbol}` for now

and also cleans up some usage of apiEndpoint in registry/api


stores/useWalletStore.tsx
```diff
@@ -40,13 +40,16 @@ import { mapFromEntries, mapEntries } from '../tools/core/script'
...
+  current: Connection
+  endpoint: string
+}
 interface WalletStore extends State {
   connected: boolean
-  connection: {
-    cluster: string
-    current: Connection
-    endpoint: string
-  }
```
jpbogle, line 49: making this object its own interface to be used elsewhere

jpbogle: ultimately i plan to make a bigger refactor to potentially unify connectionContext and rpcContext but pull conectionContext out of WalletStore into its own state so it can be used separately since not all cases require a wallet

SebastianBor: yeah, that's the right direction

PR #111, opened 2021-12-04
Feature/add treasuary account

abrzezinski94:


components/TreasuryAccount/AccountsCompactWrapper.tsx
```diff
@@ -1,27 +1,76 @@
...
           <>
-            <h3 className="mb-4">Treasury</h3>
+            <h3 className="mb-4 flex items-center">
+              Treasury
+              {!isNewAccountRoute && (
+                <Tooltip
+                  contentClassName="ml-auto"
+                  content={
+                    !canCreateGovernance &&
+                    'Connect your wallet to create new account'
```
SebastianBor, line 54: `Please connect your wallet to create a new treasury account`

Ideally we should split the tooltip messages into two states for all the buttons
1) Wallet not connected - `Please connect your wallet to create a new treasury account`
2) Governance power check - 'You don't have enough governance power to create a new treasury account' 
    The 'governance power' wording should also be generalised using a utility function because for particular use cases we can provide more details like 'not enough governance tokens', 'not enough council tokens', 'not enough governance weight' etc.

PR #116, opened 2021-12-07
try to finalize vote before withdrawing tokens

mschneider:


components/TokenBalanceCard.tsx
```diff
@@ -194,10 +199,27 @@ const TokenDeposit = ({
           // If the Proposal is in Voting state refetch it to make sure we have the latest state to avoid false positives
           proposal = await getProposal(connection, proposal.pubkey)
           if (proposal.info.state === ProposalState.Voting) {
-            // Note: It's technically possible to withdraw the vote here but I think it would be confusing and people would end up unconsciously withdrawing their votes
-            throw new Error(
-              `Can't withdraw tokens while Proposal ${proposal.info.name} is being voted on. Please withdraw your vote first`
+            const governance = await getGovernance(
```
SebastianBor, line 202: nit: There is no need to fetch governance from the chain here, you can use `{governances} = useRealm()` to get all the governances for the given realm

components/TokenBalanceCard.tsx
```diff
@@ -194,10 +199,27 @@ const TokenDeposit = ({
...
+              proposal.info.governance
             )
+            if (proposal.info.getTimeToVoteEnd(governance.info) > 0) {
+              // Note: It's technically possible to withdraw the vote here but I think it would be confusing and people would end up unconsciously withdrawing their votes
+              throw new Error(
+                `Can't withdraw tokens while Proposal ${proposal.info.name} is being voted on. Please withdraw your vote first`
+              )
+            } else {
+              // finalize proposal before withdrawing tokens so we don't stop the vote from succeeding
+              await withFinalizeVote(
```
SebastianBor, line 213: We are increasing the chance of going over the transaction size limit and the withdraw operation failing. 
I've tested that we can have up to 9 `RelinquishVote` + `Withdraw`insructions in one transaction, so maybe we could partition it? 
Could we use the multi transaction logic here? Or even if we ask the user to sign multiple times then it would be still better than silently failing on Withdraw

SebastianBor (commented):
The Finalise instruction should also be called when user calls `Release Tokens` directly from the proposal

mschneider:
Yeah we should use the multi-tx helper here :)

components/TokenBalanceCard.tsx
mschneider: added for now chunking into 8 instructions per tx, that should succeed 🤞

PR #125, opened 2021-12-09
refactor: move realm data to external JSON

johnrees:
_I know the plan is to get all the realm metadata on chain_ but in the meantime I think that moving it out of the code and into external JSON file(s)  à la [solana tokenlist](https://raw.githack.com/solana-labs/token-list/main/src/tokens/solana.tokenlist.json) could be useful for this project. If you don't agree, no worries :)

It enables a couple of things -

- external projects will be able to fetch a list of DAOs until the metadata is stored on chain, this would be particularly useful for notifications stuff that I am working on
- contributors can add their DAO without needing to go into code and potentially create conflicts with open feature branches or accidentally break something

### Images

Images aren't currently absolute URLs, they probably should be if this was merged as people might pull a versioned github URL rather than something like https://dao-beta.mango.markets/realms/mainnet-beta.json

### mainnet vs mainnet-beta?

I noticed the project has chosen to use `mainnet` over `mainnet-beta`, and its own cluster type `export type EndpointTypes = 'mainnet' | 'devnet' | 'localnet'` rather than using or extending `export type Cluster = 'devnet' | 'testnet' | 'mainnet-beta'` provided by web3.js, is there a reason for that? I'm happy to use mainnet on this project going forward if that's the preferred nomenclature

### Versioning

There's no versioning in this JSON, I just left it as a simple array to try and minimise that changes I was making here, but maybe it should have something as a root object key? e.g.

```json
{
  "version": 1,
  "realms": [
  ]
}
```

I'd also be a fan of having the realms as an object with the realmId as the key to enforce uniqueness -
```json
{
  "DPiH3H3c7t47BMxqTxLsuPQpEC6Kne8GA9VXbxpnZxFE": {
    "symbol": "MNGO",
    "displayName": "Mango DAO",
    "programId": "GqTPL6qRf5aUuqscLh8Rg2HTxPUXfhhAXDptTLhp1t2J",
  },
  "759qyfKDMMuo9v36tW7fbGanL63mZFPNbhU7zjPrkuGK": {
    "symbol": "SOCEAN",
    "programId": "5hAykmD4YGcQ7Am3N7nC9kyELq6CThAkU82nhNKDJiCy"
  }
}
```

but I understand if that's not a popular choice

### Separate files

I also split `mainnet-beta` realms and `devnet` into separate JSON files, I've never really understood why the [solana tokenlist](https://raw.githack.com/solana-labs/token-list/main/src/tokens/solana.tokenlist.json) has collated everything into a single file? _update: on reflection I guess it's to share tags etc, but it idk about the trade-off?_

I suspect the majority of devs would want to fetch all the DAOs for a single chain, but maybe I'm wrong here?

### Type checking

Typescript will check the JSON for us here, so if there is a missing required key in the JSON there will be a warning.

Here's what happened when I omitted a programId in the devnet DAO JSON for example.

Screenshot 2021-12-09 at 8 21 51 PM

If we added `yarn type-check` as a CI check it would also detect bad data and block the PR merge

Screenshot 2021-12-09 at 8 50 09 PM

### DIFF
--- a/models/registry/api.ts
+++ b/models/registry/api.ts
@@ -37,14 +37,11 @@ const MAINNET_REALMS = parseRealms(mainnetBetaRealms)
 const DEVNET_REALMS = parseRealms(devnetRealms)
 
 function parseRealms(realms: RealmInfoAsJSON[]) {
-  return realms.map(
-    (realm) =>
-      ({
-        ...realm,
-        programId: new PublicKey(realm.programId),
-        realmId: new PublicKey(realm.realmId),
-      } as RealmInfo)
-  )
+  return realms.map((realm) => ({
+    ...realm,
+    programId: new PublicKey(realm.programId),
+    realmId: new PublicKey(realm.realmId),
+  })) as ReadonlyArray<RealmInfo>
 }
 
 export function getAllRealmInfos({ cluster }: ConnectionContext) {
--- a/models/registry/api.ts
+++ b/models/registry/api.ts
@@ -1,4 +1,6 @@
 import { PublicKey } from '@solana/web3.js'
+import devnetRealms from 'public/realms/devnet.json'
+import mainnetBetaRealms from 'public/realms/mainnet-beta.json'
 import { ConnectionContext } from 'stores/useWalletStore'
 import { equalsIgnoreCase } from '../../tools/core/strings'
 
@@ -6,6 +8,7 @@ export enum ProgramVersion {
   V1 = 1,
   V2,
 }
+
 export interface RealmInfo {
   symbol: string
   programId: PublicKey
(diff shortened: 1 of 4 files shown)


SebastianBor:
> ### Images
> Images aren't currently absolute URLs, they probably should be 

 Yeah, agree, this is a good idea 

> ### mainnet vs mainnet-beta?
> I noticed the project has chosen to use `mainnet` over `mainnet-beta`, and its own cluster type 

I don't think there is a specific reason for that, we should just use Cluster from web3.js as you suggested

> 
> ### Versioning

OK

> 
> I'd also be a fan of having the realms as an object with the realmId as the key to enforce uniqueness -

I don't have a strong opinion here, but if you insist then let's do it :)

> ### Separate files

cool

> ### Type checking

yeah, let's do that

PR #126, opened 2021-12-10
feat: Marinade DAO added to devnet dashboard

MjCage:
Listed Marinade DAO entry on the devnet dashboard. Marinade team is looking to run some tests and become familiar with the new user interface.


SebastianBor:
> Listed Marinade DAO entry on the devnet dashboard. Marinade team is looking to run some tests and become familiar with the new user interface.

We moved the list to separate JSON file, please update accordingly

PR #131, opened 2021-12-13
switch to solana wallet adapter

mschneider:
this will allow:
1. easy onboarding of more wallets
2. sendTransaction**s** helper to quickly approve multiple tx
3. fixes real loading issues on devnet


ptaffet:
You missed updating a few import lines, so the code doesn't build right now:
https://github.com/blockworks-foundation/governance-ui/blob/d4f2071c0617c12f1e11d13191fa8328f1eba007/utils/ataTools.tsx#L7
https://github.com/blockworks-foundation/governance-ui/blob/d4f2071c0617c12f1e11d13191fa8328f1eba007/utils/instructionTools.ts#L9

PR #134, opened 2021-12-14
wip: Realm wizard

Mury12:
### Changelog

 - [X] Setup the Realm Wizard to automatically create realms
 - [X] Refactored the realm page to better maintenance and readability
 - [X] Single responsibility components to improve the form usability
 - [x] Added a `placeholder` prop into the `Input` component
 - [x] Added a wrapper to the Loading component to center the icon
 - [x] Tested the realm creation 
 - [x] Removed testing modifications ( some components were modified due to testing purposes )
 - [x] Solved the large transaction problem. ~The transaction with more than 3 wallets (including the current wallet) is too large to send. Splitting it into smaller chunks doesn't seem possible due to the number of signers versus the number of instructions that are created, this way, we can't split the council mint instructions into smaller chunks because it is not possible to know which signer belongs to each set of instructions.~ Chunk council transactions in sets of 10.


components/RealmWizard/RealmWizard.tsx
```diff
@@ -0,0 +1,356 @@
...
+        artifacts.communityMintId
+      )
+      artifacts = {
+        ...artifacts,
+        ...communityMintInfo,
+      }
+    }
+    setForm(artifacts)
+    setHasGeneratedArtifacts(true)
+    return artifacts
```
Mury12, line 156: remove

PR #142, opened 2021-12-15
test: fix `yarn test`

johnrees:
1. fixes `yarn test` with `next-router-mock`

2. runs github workflow to ensure all tests and linter pass (`yarn test-all`)

preview: https://github.com/johnrees/governance-ui/runs/4538845779?check_suite_focus=true

3. removes unused test snapshot

I'm planning to add some tests to this repo and this will be a good starting point for me to build from. Thanks!


johnrees:
can someone enable github actions pls?

johnrees:
the linter checks are currently in husky a pre-push hook but the issue that is that if someone pushes without installing husky, or if they make edits directly on files using github's UI then they will not run

johnrees:
alternatively the tests could run on netlify but I find that things are generally quicker and you get more flexibility with GH actions.

I'd also like to try running the notifier bot in a github action on a schedule if possible, as I think that it could make it easier for forkers to run their own.

PR #145, opened 2021-12-16
fix: remove `ProgramVersion` circular dependency

johnrees:
While setting up tests I noticed there's a circular dependency which is causing some unexpected behaviour. 

Calls to `ProgramVersion.V1` will sometimes fail because `ProgramVersion === undefined` [when there are circular dependencies with enums](https://stackoverflow.com/a/59665223)

### before

Screenshot 2021-12-16 at 10 32 21 PM

### after

Screenshot 2021-12-16 at 10 31 57 PM


johnrees:
These red nodes are the files linked to `scripts/governance-notifier.ts` which are affected by circular imports

graph

and after

graph

I think I will add the borsh fix in this PR too

johnrees:
nvm, I'll come back to `borsh` in a separate PR if it's an issue

PR #146, opened 2021-12-17
test: setup jest to run API/notification tests in isolation

johnrees:
With these foundations I'll be able to move forward and develop notification features quickly and ensure that future changes won't break existing functionality.

I've written more tests and functionality for the notifier scripts but I would like to add them in a separate PR after this one if possible in order to keep things simple to review and hopefully merge.

## Notes

* `jsdom` breaks the borsh struct parsing, so the notifier tests have to run in the `node` environment. Not entirely sure what's happening here yet, but it's an interesting discovery! It'll kind of work with `jsdom`, but returns empty strings for some fields like proposal `name`


pages/index.tsx
```diff
@@ -6,10 +6,10 @@ const Index = () => {
   useEffect(() => {
     const { REALM } = process.env
     const mainUrl = REALM ? `/dao/${REALM}` : '/realms'
-    router.push(mainUrl)
+    router.replace(mainUrl)
```
johnrees, line 9: switched to replace so it that it doesn't add an extra page in the browser history

test/setup.js
```diff
@@ -0,0 +1,3 @@
+import '@testing-library/jest-dom/extend-expect'
```
johnrees, line 1: recommended on https://nextjs.org/docs/testing

test/pages/index.test.tsx
```diff
@@ -1,10 +1,39 @@
...
+    render(<Home />)
+
+    await waitFor(() => {
+      expect(singletonRouter).toMatchObject({
+        pathname: '/dao/MNGO',
+      })
+    })
+  })
+
+  test.todo(`/realms when process.env.REALM is not a valid realm symbol`)
```
johnrees, line 38: will address this in a separate PR, just wanted to have some tests which checked that everything loaded ok

next.config.js
```diff
@@ -8,7 +8,6 @@ const withTM = require('next-transpile-modules')([
 ])
 
 module.exports = withTM({
-  target: 'serverless',
```
johnrees, line 11: next recommended removing this when i ran `yarn test` https://nextjs.org/docs/messages/deprecated-target-config

test/models/registry/api.test.ts
```diff
@@ -0,0 +1,14 @@
...
+test('getCertifiedRealmInfo', async () => {
+  const mango = realms.find(({ symbol }) => symbol === 'MNGO')!
+
+  const realmInfo = await getCertifiedRealmInfo(
+    mango.symbol,
+    getConnectionContext('mainnet')
+  )
+
+  expect(realmInfo!.realmId.toBase58()).toEqual(mango.realmId)
+})
```
johnrees, line 14: I will be replacing this test with more useful ones but this is just to ensure that I can run tests without loading any frontend/next-specific code

johnrees:
sorry about all the force-pushed & vercel deployment noise here, I'll try to avoid that in future PRs

next.config.js
SebastianBor: :+1:

pages/index.tsx
SebastianBor: :+1:

test/models/registry/api.test.ts
SebastianBor: :+1:

test/pages/index.test.tsx
SebastianBor: :+1:

test/setup.js
SebastianBor: :+1:

SebastianBor (commented):
Superb

PR #149, opened 2021-12-17
fix: `yarn notifier`

johnrees:
# Problem

`yarn notifier` was unable to run for me locally.

## issue 1 - tsconfig baseUrl/paths

I was getting an error that it cannot find module `stores/useWalletStore`

Screenshot 2021-12-17 at 7 19 03 PM
I fixed this by adding `tsconfig-paths` into the tsconfig

```diff
 "ts-node": { 
+   "require": ["tsconfig-paths/register"], 
    "compilerOptions": { 
      "module": "commonjs" 
    } 
 } 
```

I also removed `tsconfig.commonjs.json` and merged it into the `ts-node` field in `tsconfig.json`

## issue 2 - unexpected token '<'

After fixing the paths I got new errors because [there is JSX in the one of the imports used by the notifier](https://github.com/solana-labs/governance-ui/blob/dbe88225610eba62c164b6b5ee5fc8cfcf469b06/components/instructions/tools.tsx#L128)

Screenshot 2021-12-17 at 7 20 26 PM

I added `react-jsx` so that the JSX would be compiled rather than preserved. I didn't love this solution because I don't think there should be JSX loaded in a headless script that's called from node.

```diff
"compilerOptions": {
+  "jsx": "react-jsx",
  "module": "commonjs"
}
```

## issue 3 - `@solana/wallet-adapter-phantom` is not cjs

The next error was because `@solana/wallet-adapter-phantom` is only exported as esmodules

Screenshot 2021-12-17 at 7 22 23 PM

I went down various rabbitholes trying to fix this (type: module in package.json, babel presets, swc, esbuild)

# Solution 🎉 

By far the simplest approach was to [move `ConnectionContext` and `getConnectionContext` into their own file](https://github.com/solana-labs/governance-ui/commit/56bf3cce25eae5e32912408a384907900d050b94#diff-4292b743ce1254e9a8127b615960dfdc0f534ea718ee3d4368f7b1ab43b7325c), this way the notifier script doesn't import any JSX or the esm `@solana/wallet-adapter-phantom` code.

#### ⚠️ I don't know if `utils/connection.ts` is the best place for this file, or if it would be better to leave it in the store file and move the JSX and wallet-adapter code out of the store file instead?

But this works for now and fixes the immediate problem.

## Finally

I also upgraded all `eslint*` dependencies because I got a warning from `yarn test-all` but I can revert that commit if it's problematic.


package.json
```diff
@@ -51,7 +51,8 @@
     "react-markdown": "^7.0.0",
     "react-portal": "^4.2.1",
     "superstruct": "^0.15.3",
-    "ts-node": "^10.2.1",
+    "ts-node": "^10.4.0",
+    "tsconfig-paths": "^3.12.0",
```
johnrees, line 55: These should probably be in `devDependencies` but I didn't want to break stuff on heroku if it's running in `NODE_ENV=production`

SebastianBor: We use Vercel so it should be ok

SebastianBor:
> #### ⚠️ I don't know if `utils/connection.ts` is the best place for this file, or if it would be better to leave it in the store file and move the JSX and wallet-adapter code out of the store file instead?

It's better here, I think store is inherently related to UI and connection is more basic concept 
Just let's unify our utility functions under `tools` and deprecate `utils`

Thank you for all that cleanup work, the project needs it !!!

PR #150, opened 2021-12-17
globally disable `@typescript-eslint/no-non-null-assertion` rule

johnrees:
this removes a lot of code comments and all existing warnings from eslint about using `variable!`


.eslintrc.json
```diff
@@ -30,6 +30,7 @@
     "@typescript-eslint/indent": 0,
     "@typescript-eslint/member-delimiter-style": 0,
     "@typescript-eslint/ban-ts-comment": 0,
+    "@typescript-eslint/no-non-null-assertion": 0,
```
johnrees, line 33: this is the only inserted code in this PR, everything else is deleting comments

PR #155, opened 2021-12-19
refactor: batch getProgramAccounts in scripts/api

johnrees:
This batches multiple `getProgramAccounts` requests so there's no need to sleep to avoid rate limiting when iterating through different account types.

After doing some tests with MNGO on api.mainnet-beta.solana.com RPC, `yarn notifier` is taking ~300ms to check for all proposals in a realm instead of ~12s

Screenshot 2021-12-19 at 12 40 14 AM

I will be doing some more checks before opening for review.


johnrees:
closing for now in favour of #170

PR #169, opened 2021-12-22
Fix/create realm button

Mury12:
### Changelog

 - [X] Changed the position of the `create realm` button: removed the plus sign at the end of the list and transposed to the top of the page 
 
image


jpbogle (approved):
lgtm, was the main concern here that you wanted to see create realm button before connecting wallet?

PR #170, opened 2021-12-22
feat: show # open proposals for each realm in dashboard

johnrees:
### This fetches counts of open proposals for all realms in 2 batched getProgramAccount requests

Screenshot 2021-12-22 at 6 33 53 PM

GPA Req 1: get all governance account pubkeys for each realm
GPA Req 2: get all proposals for each governance account from req 1, filtered by `memcmp: { offset: 1 + 32 + 32, bytes: encode([ProposalState.Voting]) }`

## TODO

- [x] exclude proposals in `Finalizing` state?
- [x] currently there are 4 requests, 2 for each dashboard. Consider pros/cons of reducing to 2 requests.
- [x] confirm design (dashboard layout might change soon anyway)
- [x] refactor `getProposalCounts`
- [x] remove snapshot tests (useful for me while refactoring)
- [x] use project's connection settings instead of hardcoded + support devnet
- [x] consider running requests through a read-optimised cached RPC endpoint e.g. https://twitter.com/john__rees/status/1437137487688777734

## References

https://twitter.com/SolanaSail/status/1471035950780276740


pages/realms/components/RealmsDashboard.tsx
```diff
@@ -93,3 +105,9 @@ export default function RealmsDashboard({
     </div>
   )
 }
+
+const Badge: React.FC<{ count: number }> = ({ count }) => (
```
SebastianBor, line 109: nit: Can this be just a functional component? Do we need the explicit typing?

scripts/getProposalCounts.ts
```diff
@@ -0,0 +1,132 @@
...
+      'content-type': 'application/json',
+    },
+  })
+
+  const json = await res.json()
+
+  return json
+}
+
+const getGovernanceAccounts = async (realms) => {
```
SebastianBor, line 23: Can we use `getGovernanceAccounts()` here? https://github.com/solana-labs/governance-ui/blob/0eb810ca07bd90dd672fd47daa192ab9e91c094b/models/api.ts#L98

SebastianBor:
> * [ ]  exclude proposals in `Finalizing` state?

Yes, let's exclude them but please note it's UI only state and time dependent. When vote time ends and `Proposal` is not automatically tipped it's still in `Voting` state but in the UI it's displayed as `Finalising`  https://github.com/solana-labs/governance-ui/blob/0eb810ca07bd90dd672fd47daa192ab9e91c094b/components/ProposalStatusBadge.tsx#L6

pages/realms/components/RealmsDashboard.tsx
johnrees: sure :) which style do you prefer? I'll be consistent with whichever one you say

```tsx
const ConstWithReactFunctionSignature: React.FC<{ count: number }> = ({ count }) => <span>{count}</span>

const ConstWithType = ({ count }: { count: number }) => <span>{count}</span>
// ^ my slight preference, but very happy to go with any

const ConstWithoutType = ({ count }) => <span>{count}</span>

function FunctionWithType({ count }: { count: number }) {
  return <span>{count}</span>
}

function FunctionWithoutType({ count }) {
  return <span>{count}</span>
}
```

scripts/getProposalCounts.ts
johnrees: yes we can, and I am keen to re-use code 👍  the only downside of using it in its current form is that it would start making _a lot_ of individual `getProgramAccount` requests instead of a single batched request. This is ok for mango.rpcpool.com but anyone who forks this repo could have rate-limit issues with different RPC providers.

To solve this I was thinking I might refactor getGovernanceAccounts (and now getBorshProgramAccounts) to have the ability to make a batched request like I did in this refactoring https://github.com/johnrees/governance-ui/blob/optimise-scripts/scripts/api.ts#L26-L100

johnrees: this issue of loading a few batched vs lots of non-batched requests is almost solved on one of my local branches

https://github.com/johnrees/governance-ui/blob/request-batching-service-worker/public/sidekick.js

it will automatically batch `getGovernanceAccounts` and any other type of RPC request without needing to make any changes to the current code

I'll open a PR next week, but feel free to reject if not interested

Screenshot 2021-12-26 at 6 56 59 PM

pages/realms/components/RealmsDashboard.tsx
SebastianBor: I prefer your preference too :) Can we enforce it through a linter?
I would like to enable strict types checks at some point

scripts/getProposalCounts.ts
SebastianBor: This is super cool

pages/realms/components/RealmsDashboard.tsx
```diff
@@ -20,7 +21,17 @@ export default function RealmsDashboard({
...
   const { fmtUrlWithCluster } = useQueryContext()
+
   const { connected, current: wallet } = useWalletStore((s) => s)
+  const [counts, setCounts] = useState({})
+
+  useEffect(() => {
+    const fetchBadgeCounts = async () => {
+      setCounts(await getNumberOfProposalsInVotingState(realms))
+    }
+    // xxx: exclude badge counts for uncharted realms for now
```
SebastianBor, line 32: I think that's ok, being charted should come with privileges and lots of spam in the uncharted territory shouldn't slow us down

scripts/accountQueries.ts
```diff
@@ -0,0 +1,79 @@
+import {
+  getAccountTypes,
+  Governance,
+  Proposal,
+  ProposalState,
+} from '@models/accounts'
+import { GOVERNANCE_SCHEMA } from '@models/serialisation'
+import { PublicKey } from '@solana/web3.js'
+import { getConnectionContext } from '@utils/connection'
+import { AccountsQuery } from 'glasseater'
```
SebastianBor, line 10: I love that npm package name :)

scripts/accountQueries.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Fetches the number of open Proposals in for a list of Realms
+ * @example
+ * // returns { "DPi...xFE": 1, "foo...123": 4 }
+ * `await getProposalCounts(mainnetRealms)`
+ * @returns {Object} Returns dictionary of `{ [realmId]: count }`
+ */
+export const getNumberOfProposalsInVotingState = async (realms) => {
+  // 1/3) get all governance accounts for each realm
+  const governancesQuery = new AccountsQuery(GOVERNANCE_SCHEMA, Governance)
```
SebastianBor, line 37: Note: the schema should be a function which resolves based on the account type to support different version of the program. See https://github.com/solana-labs/oyster/blob/6b48d97056b789129deec46591481f0074ff823c/packages/governance/src/models/core/api.ts#L75

asktree:
closing on account of age and, as a more minor consideration, custom npm package usage.

PR #179, opened 2021-12-27
chore: enforce valid GOVERNANCE_SCHEMA keys with types

johnrees:
A mini DX update which will only allow valid classes to be used as keys when accessing `GOVERNMENT_SCHEMA`

currently anything can be used as a key because of the `Map<any,any>` type, this makes it a bit stricter and typescript will warn if trying to access a key that doesn't exist.

## before
Screenshot 2021-12-27 at 12 08 05 PM

## after
Screenshot 2021-12-27 at 12 06 25 PM


SebastianBor:
Very cool!!!

PR #181, opened 2021-12-27
Add docs

rmshea:


pindaroso:
I noticed this just add the Docusaurus framework with all its boilerplate. No actual Realms documentation.

rmshea:
yep @pindaroso working with @SebastianBor to get documentation in

pindaroso:
Ah, brilliant. LMK if you need a hand!

PR #185, opened 2021-12-28
Feat/polling wallet

ThalesBMC:
- [x] Implemented a solution to listen to Phantom wallet's account change

### Reason
Phantom wallet doesn't support an onAccountChange listener, causing confusion when the user changes his wallet.

### Solution
Use of window.solana.connect() error handling to identify if the current wallet is still connected and if not update on the website.


SebastianBor:
Fixed in https://github.com/solana-labs/governance-ui/pull/684

PR #196, opened 2021-12-29
Fix/members btn

abrzezinski94:


components/Members/MembersCompactWrapper.tsx
```diff
@@ -33,47 +35,49 @@ const MembersCompactWrapper = () => {
   const goToAddMemberView = () => {
     setCurrentCompactView(ViewState.AddMember)
   }
+  const addNewMemberTooltip = !connected
+    ? 'Connect your wallet to add new council member'
+    : !canUseMintInstruction
+    ? "You don't have enough governance power to add new council member"
+    : !canMintRealmCouncilToken()
```
SebastianBor, line 42: `canMintRealmCouncilToken()` should be checked before `canUseMintInstruction`

PR #202, opened 2021-12-30
Feature/look for members without dep tokens

abrzezinski94:


scripts/api.ts
```diff
@@ -133,3 +135,22 @@ async function getGovernanceAccountsImpl<TAccount extends GovernanceAccount>(
   }
   return accounts
 }
+
+export const getTokenAccountsByMint = async (
```
SebastianBor, line 139: This function should be in `tokens.tsx`  and return parsed accounts. See `getOwnedTokenAccounts` as an example

PR #219, opened 2022-01-04
fix: footer responsiveness on smaller screens

ThalesBMC:
- [x] made the footer responsive on smaller screens, according to design.


abrzezinski94:
looks good

PR #222, opened 2022-01-04
Hotfix/bespoke form defaults

Mury12:
### Changelog
 
 - [x] Fixed form not showing default values for Min community tokens to create governance and community supply factor
 - [x] Refactored the ATA creation and council token mint to avoid duplicated mints and accounts.
 - [x] Labels informing the percentages 
 - [x] Fixed the wrong MintMaxVoteWeightSource to be shifted by the decimals
 - [x] Always show the min community tokens to create and community mint supply factor


actions/registerRealm.ts
```diff
@@ -41,14 +49,37 @@ import { withSetRealmAuthority } from '@models/withSetRealmAuthority'
...
-const COMMUNITY_MINT_DECIMALS = 6
+export const COMMUNITY_MINT_DECIMALS = 6
+
+/**
+ * Try to get the associated token account of an owner to a mint id
+ * @param connection
+ * @param owner
+ * @param mint
+ */
+async function tryGetAta(
```
SebastianBor, line 67: Why is this function duplicated here? Can we use this one https://github.com/solana-labs/governance-ui/blob/2639a59f203d274bbb3a81361fab69b630aa274b/utils/validations.tsx#L48 ?

It really should be moved to tokens.tsx to be consistent

actions/registerRealm.ts
```diff
@@ -97,15 +128,27 @@ async function prepareMintInstructions(
...
+          connection,
           ownerPk,
-          walletPubkey
+          _mintPk
         )
+        const shouldMint = !ata?.account.amount.gt(new BN(0))
+
+        const ataPk =
+          ata?.publicKey ??
+          (await withCreateAssociatedTokenAccount(
```
SebastianBor, line 140: If the Ata was created before the this instruction would fail, you have to check for Ata existence separately, and if it exists then skip this instruction

actions/registerRealm.ts
Mury12: fixed

components/RealmWizard/RealmWizard.tsx
```diff
@@ -143,31 +142,36 @@ const RealmWizard: React.FC = () => {
...
+            walletPubkey: wallet!.publicKey!,
+          },
+          new PublicKey(form.governanceProgramId!),
+          form.programVersion ?? ProgramVersion.V1,
+          form.name!,
+          form.communityMintId
+            ? new PublicKey(form.communityMintId)
+            : undefined,
+          form.councilMintId ? new PublicKey(form.councilMintId) : undefined,
+          MintMaxVoteWeightSource.FULL_SUPPLY_FRACTION,
```
Mury12, line 159: Static value instead of form's value

PR #220, opened 2022-01-04
Feature/nft treasuary

abrzezinski94:

PR #224, opened 2022-01-05
feat/treasury payment quick action

amandagonsalves:


components/TreasuryAccount/TreasuryPaymentForm.tsx
```diff
@@ -0,0 +1,471 @@
...
+  const { fetchRealmGovernance } = useWalletStore((s) => s.actions)
+
+  const tokenInfo = useTreasuryAccountStore((s) => s.compact.tokenInfo)
+  const wallet = useWalletStore((s) => s.current)
+  const router = useRouter()
+  const programId: PublicKey | undefined = realmInfo?.programId
+
+  const [form, setForm] = useState<SendTokenCompactViewForm>({
+    destinationAccount: '',
+    amount: 0,
```
SebastianBor, line 68: There shouldn't be default amount, and amount 0 should be accepted in case somebody wants to do dry run proposal

SebastianBor (commented):
In the context view we can see the treasury account we make the transfer from, but in the modal view it's lost, so it's not possible for example to tell how much it's there, Please add it to the modal view 

image

The treasury info is missing
image

SebastianBor:
When user has too many outstanding proposals then the Send button (in the side view) should be disabled with the corresponding tooltip. Note both council and community proposals should be checked, see https://github.com/solana-labs/governance-ui/blob/d9ad2e20827c71e6772965273e9308de14942bd7/components/AssetsList/AssetsCompactWrapper.tsx#L35 for reference which has similar logic

SebastianBor:
Most of this work has been done by @abrzezinski94 as part of the NFT treasury PR https://github.com/solana-labs/governance-ui/pull/220. Please sync with master and only fix the issues which still remains

PR #246, opened 2022-01-12
Feat/transaction status component

ThalesBMC:
- [x] Created component according to design to show transaction result at deposit/withdraw and cancel/finalize/sign off transactions.
- [x] The loader is now at the new component 
- [x] Removed the result of the transaction from notification, and now it appears in the new component.
- [x] Fixed css of proposalStateBadge and text that was overflowing.
- [x] Added see details options so you can see the transaction details


SebastianBor:
> @SebastianBor as far as I know we currently have a transaction status component that gets used everywhere. I am not sure of its relationship to this PR.

yes, we do, this PR was a different take on that problem

PR #271, opened 2022-01-21
Feat/send txn widget

Mury12:
### Changelog

 - [X] Implement a component to handle send transactions with a progress bar, transaction names and step names.
 - [X] Returns only high level errors
 - [X] Returns all the `txIds` when `onFinish` is called 

Applied to Create DAO page

Recording 2022-01-21 at 18 09 02


asktree:
closing because it is a 2yo draft

PR #284, opened 2022-01-25
Add cardinal twitter link popup

jpbogle:
Wanted to start getting feedback on this

I wasnt sure exactly how to display the buttons, and if it should still disconnect when you click on the wallet button, for now i disabled that and put it in the menu itself along with the link twitter button

It also should work for members here in the MemberItem I think I still need to add it to MemberOverview

image

image

image

It also should work for members here

image


jpbogle:
@abrzezinski94 @SebastianBor

PR #291, opened 2022-01-31
Feature/native sol

abrzezinski94:


components/instructions/programs/splToken.tsx
```diff
@@ -30,6 +31,50 @@ export function getMintMetadata(
     : MINT_METADATA[tokenMintAddress]
 }
 
+export const SYSTEM_INSTRUCTIONS = {
```
SebastianBor, line 34: let's move it to `system.tsx`

PR #307, opened 2022-02-04
lock token fixes

abrzezinski94:


VoteStakeRegistry/actions/getGrantInstruction.ts
```diff
@@ -86,9 +84,10 @@ export const getGrantInstruction = async ({
         voterWeightRecord: voterWeightPk,
         vault: voterATAPk,
         depositToken: fromPk,
-        authority: grantAuthority,
-        payer: feePayerPk,
+        tokenAuthority: grantAuthority,
```
SebastianBor, line 87: This is governance right? `grantAuthority` is confusing name here because it can be any governance

Issue #312, opened 2022-02-05
Disable 'Send It' button for Proposal comment when user has no voting power and show tooltip


SebastianBor:

RaymondDashWu:
About to make a PR for this but [this line](https://github.com/solana-labs/governance-ui/blob/b9852885d582074522412b1378e7fe7020499ec1/VoteStakeRegistry/actions/getGrantInstruction.ts#L87) is preventing me from pushing. I haven't touched that part of the codebase yet. Any advice?

```
Argument of type '[number, number, { [x: string]: {}; }, BN, number, boolean, BN, { accounts: { registrar: PublicKey; voter: PublicKey; voterAuthority: PublicKey; ... 10 more ...; rent: PublicKey; }; }]' is not assignable to parameter of type '[...unknown[], Context<Accounts<{ name: "registrar"; isMut: false; isSigner: false; } | { name: "voter"; isMut: true; isSigner: false; } | { name: "voterAuthority"; isMut: false; isSigner: false; } | { name: "voterWeightRecord"; isMut: true; isSigner: false; } | ... 8 more ... | { ...; }>>]'.
  Type at position 7 in source is not compatible with type at position 1 in target.
    The types of 'accounts' are incompatible between these types.
      Type '{ registrar: PublicKey; voter: PublicKey; voterAuthority: PublicKey; voterWeightRecord: PublicKey; vault: PublicKey; depositToken: PublicKey; ... 7 more ...; rent: PublicKey; }' is not assignable to type 'Accounts<{ name: "registrar"; isMut: false; isSigner: false; } | { name: "voter"; isMut: true; isSigner: false; } | { name: "voterAuthority"; isMut: false; isSigner: false; } | { name: "voterWeightRecord"; isMut: true; isSigner: false; } | ... 8 more ... | { ...; }>'.
        Object literal may only specify known properties, and 'tokenAuthority' does not exist in type 'Accounts<{ name: "registrar"; isMut: false; isSigner: false; } | { name: "voter"; isMut: true; isSigner: false; } | { name: "voterAuthority"; isMut: false; isSigner: false; } | { name: "voterWeightRecord"; isMut: true; isSigner: false; } | ... 8 more ... | { ...; }>'.ts(2345)

```

PR #315, opened 2022-02-06
'Send it' button disabled when users have no voting power

RaymondDashWu:
Solves https://github.com/solana-labs/governance-ui/issues/312

# Changelog
- Users can no longer submit proposal comments when they have no voting power
- Button is disabled with tooltip 'You need to have voting power for this community to submit your comment.'


RaymondDashWu:
Logic issue. Have to know what voter weight is comes from voter weight add in or Time based (example MANGO). @abrzezinski94 comments?

PR #343, opened 2022-02-12
UXD fork back -  Solend + Multi Itx support + Create ATA itx

ThomasProust:
First Part of the Governance-UI fork back from UXD Protocol into Solana-Labs.
Features include:

- Update UXP Governance infos
- Update Anchor version to ^0.20.1
- Add Solend instructions through Solend SDK
- Add Multi-Instruction per proposal support
- Add Execute All instructions for a Proposal

Capture d’écran 2022-02-12 à 23 42 50
Capture d’écran 2022-02-12 à 23 42 08
Capture d’écran 2022-02-12 à 23 32 33
Capture d’écran 2022-02-12 à 23 31 55
Add Create Associated Token Account Instruction


mschneider:
@ThomasProust is there a specific reason you need an instruction to create ATA's? we usually just create them in preflight instructions

hooks/useGovernanceAssets.ts
```diff
@@ -162,6 +164,41 @@ export default function useGovernanceAssets() {
         canUseTokenTransferInstruction &&
         realm?.account.config.useCommunityVoterWeightAddin,
     },
+    {
+      id: Instructions.CreateAssociatedTokenAccount,
+      name: 'Create Associated Token Account',
+      isVisible: canUseAnyInstruction,
+    },
```
mschneider, line 171: is there a specific reason you need an instruction to create ATA's? we usually just create them in preflight instructions

hooks/useGovernanceAssets.ts
```diff
@@ -162,6 +164,41 @@ export default function useGovernanceAssets() {
...
+    {
+      id: Instructions.CreateSolendObligationAccount,
+      name: 'Solend: Create Obligation Account',
+      isVisible: canUseAnyInstruction,
+    },
```
mschneider, line 176: please use correct visibility helpers, i assume that some of the following instructions should only be visible if users have access to token transfers etc.

hooks/useGovernedMultiTypeAccounts.ts
```diff
@@ -0,0 +1,68 @@
+import { GovernedMultiTypeAccount } from '@utils/tokens'
+import { useCallback, useEffect, useState } from 'react'
+import useGovernanceAssets from './useGovernanceAssets'
+
+export default function useGovernedMultiTypeAccounts() {
```
mschneider, line 5: i dont get what this is about without reading the code, could this be useGovernedTokenAndMintAccounts?

pages/dao/[symbol]/proposal/components/instructions/Solend/WithdrawObligationCollateralAndRedeemReserveLiquidity.tsx
```diff
@@ -0,0 +1,180 @@
...
+        onChange={(value) => {
+          handleSetForm({ value, propertyName: 'governedAccount' })
+        }}
+        value={form.governedAccount}
+        error={formErrors['governedAccount']}
+        shouldBeGoverned={shouldBeGoverned}
+        governance={governance}
+      />
+
+      <Select
```
Comment (mschneider) [inline pages/dao/[symbol]/proposal/components/instructions/Solend/WithdrawObligationCollateralAndRedeemReserveLiquidity.tsx:149]: token is implicit through obligation

hooks/useGovernanceAssets.ts
ThomasProust: I think we had the case where we wanted the ATA creation to be visible in the proposal, but can't recall for sure. Would you rather see that instruction removed?

hooks/useGovernedMultiTypeAccounts.ts
ThomasProust: This allows us to select any type of governance (program/account/mint/token), so what about `useGovernedAllTypeAccounts`?

pages/dao/[symbol]/proposal/components/instructions/Solend/DepositReserveLiquidityAndObligationCollateral.tsx
```diff
@@ -0,0 +1,179 @@
...
+      .number()
+      .moreThan(0, 'Amount should be more than 0')
+      .required('Amount is required'),
+  })
+
+  return (
+    <>
+      <GovernedAccountSelect
+        label="Governance"
+        governedAccounts={governedMultiTypeAccounts}
```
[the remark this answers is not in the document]
Comment (ThomasProust) [reply pages/dao/[symbol]/proposal/components/instructions/Solend/DepositReserveLiquidityAndObligationCollateral.tsx:138]: For this instruction we are leveraging ATAs, so in theory any kind of governed account can be used as long as it has the corresponding ATAs created.

PR #344, opened 2022-02-13
Toggle Between Mainnet/Devnet

RaymondDashWu:
# Changelog
- Added a toggle to switch between mainnet/devnet 
image
image
image
image

### DIFF
--- a/components/ConnectWalletButton.tsx
+++ b/components/ConnectWalletButton.tsx
@@ -1,4 +1,4 @@
-import { Menu, Switch } from '@headlessui/react'
+import { Menu } from '@headlessui/react'
 import { useMemo, useState } from 'react'
 import { CheckCircleIcon, ChevronDownIcon } from '@heroicons/react/solid'
 import styled from '@emotion/styled'
@@ -16,7 +16,9 @@ import {
 import { BackspaceIcon } from '@heroicons/react/solid'
 import { UserCircleIcon } from '@heroicons/react/outline'
 import { abbreviateAddress } from '@utils/formatting'
+import { useRouter } from 'next/router'
 import TwitterIcon from './TwitterIcon'
+import Switch from './Switch'
 
 const StyledWalletProviderLabel = styled.p`
   font-size: 0.65rem;
@@ -37,6 +39,15 @@ const ConnectWalletButton = (props) => {
   ])
 
   const [useDevnet, setUseDevnet] = useState(false)
+  const router = useRouter()
+  const handleToggleDevnet = () => {
+    setUseDevnet(!useDevnet)
+    if (useDevnet) {
+      router.push(`${window.location.pathname}`)
+    } else {
+      router.push(`${window.location.href}?cluster=devnet`)
+    }
+  }
 
   const handleConnectDisconnect = async () => {
     try {
@@ -185,13 +196,19 @@ const ConnectWalletButton = (props) => {
                           <span className="text-sm">Disconnect</span>
                         </button>
                       </Menu.Item>
-                      <Menu.Item key={'devnet'} onClick={() => {}}>
+                      <Menu.Item
+                        key={'devnet'}
+                        onClick={() => {
+                          handleToggleDevnet()
+                        }}
+                      >
                         <button className="flex default-transition h-9 items-center p-2 w-full hover:bg-bkg-3 hover:cursor-pointer hover:rounded font-normal focus:outline-none">
                           <span className="text-sm">Devnet</span>
                           <Switch
-                            className="LOOKHERE"
                             checked={useDevnet}
-                            onChange={setUseDevnet}
+                            onChange={() => {
+                              handleToggleDevnet()
+                            }}
                           />
                         </button>
                       </Menu.Item>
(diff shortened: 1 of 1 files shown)


components/ConnectWalletButton.tsx
```diff
@@ -36,6 +38,17 @@ const ConnectWalletButton = (props) => {
...
   ])
 
+  const [useDevnet, setUseDevnet] = useState(false)
+  const router = useRouter()
+  const handleToggleDevnet = () => {
+    setUseDevnet(!useDevnet)
+    if (useDevnet) {
+      router.push(`${window.location.pathname}`)
+    } else {
+      router.push(`${window.location.href}?cluster=devnet`)
```
0xCryptoSheik, line 48: Hi @RaymondDashWu @SebastianBor @abrzezinski94

A quick comment about this feature.
I strongly believe that such a switch should trigger a full page reload.
It would have multiple benefits relative to the maintainability, stability of the app, and would allow for some simplifications in the state management.

This is something that I have previously advocated for & implemented in Metaplex for example: https://github.com/metaplex-foundation/metaplex/pull/958 as well as in UXD Protocol's app.

In the case of UXD Protocol, this has allowed us to remove notion of active cluster & connection state handling altogether, and as a result, significantly simplify the codebase.
As you can see in https://app.uxd.fi, the `connection` is available globally in the app from `window.__UXD__.connection`, as well as other things which do not change over the app's lifecycle, and as such do not need to be stored within a state object, managed by Redux or anything else.

One parallel that works pretty well I believe is to think about the connection object or active cluster the same way we usually think about an active locale / translation strings in an app, it's indeed very common to trigger a full page reload when the user changes such a central preference, which allows for a much simpler implementation overall.

Happy to discuss this further, I can create an issue if you think that's more appropriate.
Thank you for your consideration.

0xCryptoSheik: Seeing this in our PRs break my heart every time:

```ts
  const connection = useWalletStore((s) => s.connection)
```

SebastianBor: > Seeing this in our PRs break my heart every time:
> 
> ```ts
>   const connection = useWalletStore((s) => s.connection)
> ```

@RaymondDashWu, @abrzezinski94  let's make sure @0xCryptoSheik heart doesn't fall apart

RaymondDashWu: @0xCryptoSheik could you go more into detail on how it's implemented in https://app.uxd.fi/? Either that or provide another example? I'm not able to see the site in the US.

image

RaymondDashWu: > Seeing this in our PRs break my heart every time:
> 
> ```ts
>   const connection = useWalletStore((s) => s.connection)
> ```

I'm also not immediately sure what's wrong with this line. Could you clarify? I'm new to the codebase.

0xCryptoSheik: In UXD's app, the cluster / connection handling is like so:

```ts
  // in app initialization, before first render
  const params = getParams(); // retrieve potential ?cluster setting from querystring
  const storage = getStorage(); // retrieve potential `cluster` setting from local storage
  ...
  export function initConfig(params: Params, storage: Storage) {
    let config = getConfigByCluster(DEFAULT_CLUSTER);

    for (const source of [params, storage]) {
      if (source.cluster !== null) {
        try {
          config = getConfigByCluster(source.cluster as ClusterName);
          break;
        } catch {
          if (source === storage) {
            // cluster from storage is invalid, clear it.
            window.localStorage.removeItem("cluster");
          }
        }
      }
    }

    return config;
  }

  // initialize appropriate config depending on cluster
  const config = initConfig(params, storage);

  // retrieve RPC endpoint
  const clusterRPCEndpoint = await initClusterRPCEndpoint(config);
  // initialize solana web3 connection 
  const connection = initConnection(clusterRPCEndpoint);
  // centralize everything web3 in a singleton, which holds the solana web3 connection
  const service = await UXDClientService.initialize(config, connection);

  ...
  
  // hold solana web3 connection & service singleton in a global object
  // allows direct usage of either directly from anywhere in the app
  window.__UXD__ = { config, service, connection, ... };
  
  ... 
  
  // switching from one cluster to another triggers a full page reload
  <a
    key={`cluster-${cluster}`}
    href={`/?cluster=${cluster}`}
    onClick={() => {
      window.localStorage.setItem("cluster", cluster);
    }}
  >
    {cluster}
  </a>
  
  // Potentially accessing the connection from anywhere in the app
  window.__UXD__.connection.getTokenAccountBalance(...)
  
  // Accessing the current cluster from anywhere in the app
  const currentCluster = window.__UXD__.config.GetClusterName(); // devnet | mainnet-beta
```


> > Seeing this in our PRs break my heart every time:
> > ```ts
> >   const connection = useWalletStore((s) => s.connection)
> > ```
> 
> I'm also not immediately sure what's wrong with this line. Could you clarify? I'm new to the codebase.

It's kind of a detail compared to my main point, but what I dislike is that the Solana/web3 connection is retrieved:
- from a state management hook (whereas we could make it so the cluster/connection never changes during the app's lifecycle)
- incorrectly named `useWalletStore`
- it would at least be better if it was retrived from a `useConnection` hook, and the underlying connection could be stord in a react ref, or a context object, if not a global variable

RaymondDashWu:
@0xCryptoSheik Finished the issue with the toggle. Please check https://github.com/solana-labs/governance-ui/pull/381. Also, could you make a separate issue for the useWalletStore? I see that it's used a lot throughout this codebase so it should be its own PR.

PR #348, opened 2022-02-16
Improve vote results display

saml33:
Added more details to the results panel and states depending on the status of the proposal.


mschneider:
somehow the yes/no vote result ratio never finishes loading here

Screen Shot 2022-02-16 at 12 00 02 PM

saml33:
Yeah, I noticed that. Wasn't doing that earlier today @mschneider

PR #373, opened 2022-02-22
Error message for non Anchor verified program

RaymondDashWu:
# Changelog
- Error message now shows when a program is not Anchor verified

image


SebastianBor:
We can't limit creating program governance to Anchor verified programs only.

PR #381, opened 2022-02-25
Devnet Toggle Fixes

RaymondDashWu:
Handing off this PR to the next person that wants to contribute as my fellowship has ended

Continuation of #344 fixing issues pointed out by [0xCryptoSheik](https://github.com/0xCryptoSheik)

# Changelog
- Devnet toggle now available whether or not wallet is connected
- Toggle now reloads page onClick
- Mainnet/Devnet state now stateless. In localStorage

# TODO
- ~~External explorer links take you to mainnet even if you're on devnet cluster~~ Handled in #626 
- Put cluster status into global variable (possibly useWalletStore)


0xCryptoSheik:
Hey @RaymondDashWu, thanks for this I'll take a look today

0xCryptoSheik:
Quick comments:
- no `history.pushState` manipulation should be needed
- the full page reload should actually be setting a new `location.href` directly
- the `cluster` local storage value should only be retrieved once on initialization (so not in a hook called on every consumer component render)
- the `cluster` querystring parameter value should be read once on initalization

0xCryptoSheik:
- the current cluster should be accessible globally by other means than ready the local storage, the local storage may not be accessible for writing but switching from devnet / mainnet should still work (just not be persisted to the next visit)

0xCryptoSheik (changes requested):
See comments

0xCryptoSheik:
On app initialization, the priority for sources of cluster value should be:
- querystring > storage > default

RaymondDashWu:
- [X] no `history.pushState` manipulation should be needed
- [X] the full page reload should actually be setting a new `location.href` directly

I'm unsure what you mean with your other comments below. Should this be done with the wallet store instead? I think that fits your other comments based on what I know about the codebase so far.
> * the `cluster` local storage value should only be retrieved once on initialization (so not in a hook called on every consumer component render)
> * the `cluster` querystring parameter value should be read once on initalization

0xCryptoSheik:
Related to https://github.com/solana-labs/governance-ui/issues/433

components/ConnectWalletButton.tsx
```diff
@@ -38,14 +38,13 @@ const ConnectWalletButton = (props) => {
...
-  const router = useRouter()
+  const [useDevnet, setUseDevnet] = useLocalStorageState('false')
   const handleToggleDevnet = () => {
     setUseDevnet(!useDevnet)
     if (useDevnet) {
-      router.push(`${window.location.pathname}`)
+      window.location.href = `${window.location.pathname}`
     } else {
-      router.push(`${window.location.href}?cluster=devnet`)
+      window.location.href = `${window.location.href}?cluster=devnet`
```
0xCryptoSheik, line 47: This is problematic, because it will result in an invalid URL if there's already one querystring parameter in the current URL.

components/ConnectWalletButton.tsx
```diff
@@ -38,14 +38,13 @@ const ConnectWalletButton = (props) => {
...
   ])
 
-  const [useDevnet, setUseDevnet] = useState(false)
-  const router = useRouter()
+  const [useDevnet, setUseDevnet] = useLocalStorageState('false')
   const handleToggleDevnet = () => {
     setUseDevnet(!useDevnet)
     if (useDevnet) {
-      router.push(`${window.location.pathname}`)
+      window.location.href = `${window.location.pathname}`
```
0xCryptoSheik, line 45: When selected from this toggle button, the cluster should always be present in the URL to assert priority over possible existing local storage preference.

components/ConnectWalletButton.tsx
```diff
@@ -163,7 +162,17 @@ const ConnectWalletButton = (props) => {
                       </button>
                     </Menu.Item>
                   ))}
-
+                  <Menu.Item key={'devnet'}>
```
0xCryptoSheik, line 165: Why change the location of the toggle inside the menu?
Unrelated to current scope

RaymondDashWu: That was a fix that Sebastian wanted. Otherwise, devnet toggle is only available when a wallet is connected

0xCryptoSheik:
> I'm unsure what you mean with your other comments below. Should this be done with the wallet store instead? I think that fits your other comments based on what I know about the codebase so far.
> 
> > * the `cluster` local storage value should only be retrieved once on initialization (so not in a hook called on every consumer component render)
> > * the `cluster` querystring parameter value should be read once on initialization

This should be done in the same PR.
I don't know much about the codebase either, maybe @SebastianBor or @abrzezinski94 can guide further or take over.

The point is that we need some function to be responsible for the app initialization, inside which the current active cluster is determined based on potential querystring & localstorage values, before making it available to the rest of the app in some way (through a global variable, or React Context, or store..).

components/ConnectWalletButton.tsx
0xCryptoSheik: Ok :+1:

components/ConnectWalletButton.tsx
RaymondDashWu: Agreed. Will push an update to fix this.

components/ConnectWalletButton.tsx
RaymondDashWu: I'm not sure I understand what you mean. The part of the codebase you're pointing to is when there isn't a cluster specified (ie mainnet). The part that's confusing me is what you mean by priority. Are you saying it should be more explicit? For example:

```
window.location.href = `${window.location.pathname}?cluster=mainnet`
```

RaymondDashWu:
> > I'm unsure what you mean with your other comments below. Should this be done with the wallet store instead? I think that fits your other comments based on what I know about the codebase so far.
> > > * the `cluster` local storage value should only be retrieved once on initialization (so not in a hook called on every consumer component render)
> > > * the `cluster` querystring parameter value should be read once on initialization
> 
> This should be done in the same PR. I don't know much about the codebase either, maybe @SebastianBor or @abrzezinski94 can guide further or take over.
> 
> The point is that we need some function to be responsible for the app initialization, inside which the current active cluster is determined based on potential querystring & localstorage values, before making it available to the rest of the app in some way (through a global variable, or React Context, or store..).

Ah yup that does sound like what I was suggesting earlier. There's a global store used [here](https://github.com/solana-labs/governance-ui/blob/daf7cdab6aea5ac75da713ed574c25f2da75229f/components/ConnectWalletButton.tsx#L28) called useWalletStore. I'll look into this as I'm a bit hesitant to add to a store when my additions serve a limited purpose.

abrzezinski94:
Hello @RaymondDashWu @0xCryptoSheik I really like the idea to make whole connection to be global variable i can take it over and refractor ui for it but i can't do it right now i can probably fit it in my backlog next month.

RaymondDashWu:
> Hello @RaymondDashWu @0xCryptoSheik I really like the idea to make whole connection to be global variable i can take it over and refractor ui for it but i can't do it right now i can probably fit it in my backlog next month.

No worries @abrzezinski94 . I can take the task as it seems like the majority of the work is done. Just a small refactor. If you're alright with it being a global variable, I'll put it in useWalletStore.

PR #386, opened 2022-02-26
swap approval quorum for vote progress if council vote

saml33:


asktree:
closing bc age and redundancy with work that followed

PR #398, opened 2022-02-28
nfts for ukraine dao

abrzezinski94:


components/TreasuryAccount/DepositNFTFromWallet.tsx
```diff
@@ -37,7 +37,7 @@ const DepositNFTFromWallet = ({ additionalBtns }: { additionalBtns?: any }) => {
     setIsLoading(true)
     setSendingSuccess(false)
     try {
-      const governance = currentAccount!.governance!.pubkey
+      const governance = currentAccount!.token!.account.owner
```
SebastianBor, line 40: let's rename to owner so it's not confused with governance

components/TreasuryAccount/SendTokens.tsx
```diff
@@ -229,7 +229,7 @@ const SendTokens = () => {
         proposalAddress = await createProposal(
           rpcContext,
           realm,
-          selectedGovernance.pubkey,
```
SebastianBor, line 232: this is wrong, we need governance here and not. the token owner. It might be the same but not necessarily

PR #405, opened 2022-03-03
fix main deposits naming + fix vote mult calculation

abrzezinski94:


VoteStakeRegistry/components/Account/DepositCard.tsx
```diff
@@ -185,11 +181,10 @@ const DepositCard = ({ deposit }: { deposit: DepositWithMintAccount }) => {
...
             <CardLabel
               label="Vote Multiplier"
-              value={calcMintMultiplier(
-                deposit.lockup.endTs.sub(deposit.lockup.startTs).toNumber(),
-                communityMintRegistrar,
-                realm
-              )}
+              value={(
+                deposit.votingPower.toNumber() /
+                deposit.currentlyLocked.toNumber()
```
ckamm, line 186: Hmh. Voting power will be for the full number of tokens in the deposit, including the vested amount. Example: `currentlyLocked = 0, unlocked = 100, votingPower = 100, votingPowerBaseline = 100`

The key question is whether we consider the unlocked tokens part of the deposit entry. And I think we should. That would mean:
- display currentLocked + currentUnlocked in the header
- vote multiplier = voting power / voting power baseline

Another approach would be to not consider the unlocked tokens part of the deposit entry. In that case
- vote multiplier = (deposit.votingPower - depositFactor * unlocked) / deposit.currentlyLocked

PR #467, opened 2022-03-18
removal of "is web3 browser"

nicetransition:
This logic works locally, however it is buggy on Vercel (prod). Removing this to prevent issues


nicetransition:
Wrong PR location 🤦🏻‍♂️

PR #476, opened 2022-03-23
Param Page Additions

masayafunakoshi1:
Added new tracking tab, separating `proposal count` info into there.

Commented out `proposal cool-off time` as it isn't needed for now (-Sebastian)

https://user-images.githubusercontent.com/59552589/159812121-907a2202-50f0-4f81-bd1d-e73a1aa54bce.mp4

### DIFF
--- a/pages/dao/[symbol]/params/index.tsx
+++ b/pages/dao/[symbol]/params/index.tsx
@@ -234,7 +234,7 @@ const Params = () => {
                   <Tabs
                     activeTab={activeTab}
                     onChange={(t) => setActiveTab(t)}
-                    tabs={['Params', 'Accounts', 'Tracking']}
+                    tabs={['Params', 'Accounts', 'Statistics']}
                   />
                 ) : null}
                 {(() => {
@@ -476,7 +476,7 @@ const Params = () => {
                         })}
                       </div>
                     )
-                  } else if (activeTab === 'Tracking') {
+                  } else if (activeTab === 'Statistics') {
                     return (
                       <>
                         <DisplayField
--- a/pages/dao/[symbol]/params/index.tsx
+++ b/pages/dao/[symbol]/params/index.tsx
@@ -234,7 +234,7 @@ const Params = () => {
                   <Tabs
                     activeTab={activeTab}
                     onChange={(t) => setActiveTab(t)}
-                    tabs={['Params', 'Accounts', 'Statistics']}
+                    tabs={['Params', 'Accounts', 'Tracking']}
                   />
                 ) : null}
                 {(() => {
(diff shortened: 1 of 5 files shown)


pages/dao/[symbol]/params/index.tsx
```diff
@@ -232,252 +232,265 @@ const Params = () => {
...
-                    />
-                    {communityMint && (
-                      <DisplayField
-                        label="Min community tokens to create a proposal"
-                        padding
-                        val={fmtMintAmount(
-                          mint,
-                          activeGovernance.account.config
-                            .minCommunityTokensToCreateProposal
+                {(() => {
```
Comment (abrzezinski94) [inline pages/dao/[symbol]/params/index.tsx:238]: Hi so now if we have more tabs it would be good to use guard clauses and move it to component.

Comment (masayafunakoshi1) [reply pages/dao/[symbol]/params/index.tsx:238]: Thanks for the feedback, I'll change them to guard clauses!

By this, "move it to component instand of using function", did you mean I should create separate components for each of the tabs and render them in? Or keep what I did, but using guard clauses instead?

Comment (abrzezinski94) [reply pages/dao/[symbol]/params/index.tsx:238]: Yes it would be good to put it in 3 separate components like AccountsView, paramsView, statsView and use guard clauses to render them.

Comment (masayafunakoshi1) [reply pages/dao/[symbol]/params/index.tsx:238]: Got it, where would you want me to store the components? With separate .tsx files in the same params folder, or make the components directly within the index.tsx file?

Comment (abrzezinski94) [reply pages/dao/[symbol]/params/index.tsx:238]: Put it in separated files.

Comment (masayafunakoshi1) [reply pages/dao/[symbol]/params/index.tsx:238]: How's this, I created separate components in a "components" folder within "params". Then used conditionals this way. Unless you prefer I use `if()` statements?
guardClauseChanges

Comment (masayafunakoshi1) [reply pages/dao/[symbol]/params/index.tsx:238]: Hmm, seems like there are changes being made right now on the params page. I'll just wait for all of them to prevent conflict

Comment (abrzezinski94) [reply pages/dao/[symbol]/params/index.tsx:238]: > How's this, I created separate components in a "components" folder within "params". Then used conditionals this way. Unless you prefer I use `if()` statements? guardClauseChanges

all good now

> Hmm, seems like there are changes being made right now on the params page. I'll just wait for all of them to prevent conflict

Its finished one small change was needed.

Comment (masayafunakoshi1) [reply pages/dao/[symbol]/params/index.tsx]: Hmm looks like there was some conflict, seems like it's from the DisplayField component. I'll try and fix it, but I was wondering if it might be better to separate DisplayField into it's own component too? 

Since it's used by the `AccountsView` `ParamsView` and `StatsView` components as well

Comment (abrzezinski94) [reply pages/dao/[symbol]/params/index.tsx]: Change was to only match program with its known name we can leave display field as its now.

Comment (masayafunakoshi1) [reply pages/dao/[symbol]/params/index.tsx]: I must've messed something up with this branch then. Seems like it's not passing the tests so I'll close this branch/PR and make a new one

PR #617, opened 2022-04-23
Update devnet.json for LeoDao

sjfisch07:
Update devnet.json for LeoDao

### DIFF
--- a/public/realms/devnet.json
+++ b/public/realms/devnet.json
@@ -125,5 +125,14 @@
     "ogImage": "https://media.discordapp.net/attachments/933224453326909490/952347955157950504/alpha-illuminati-logo.png",
     "website": "https://www.alphailluminati.xyz",
     "twitter": "@illuminatialpha"
+  },
+  {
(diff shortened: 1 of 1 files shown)


public/realms/devnet.json
```diff
@@ -128,7 +128,7 @@
   },
   {
     "symbol": "LEO",
-    "displayName": "LeoDao New York City",
+    "displayName": "LeoDaoNewYorkCity _ DEMO",
     "programId": "GovER5Lthms3bLBqWub97yVrMmEogzX7xNjdXpPPCVZw",
```
SebastianBor, line 132: the DAO is created using `GTesTBiEWE32WHXXE2S4XbZvA5CrEc4xs6ZgRe895dP` instance

sjfisch07: Okay thanks,. Was not aware you could not change that through the JSON. Will update/create a new request

sjfisch07: corrected the programId and pushed it again.

PR #632, opened 2022-04-26
Delegate council/community token to a wallet

ndrabins:
- Shows delegated wallet in member overview under the token
- In account page, can set a wallet to be delegated council/community tokens
- Can switch between delegated accounts for community/council tokens to vote/create proposal/comment as that member
- Delegates card that shows all delegates of currently connected wallet


VoteStakeRegistry/components/Account/Account.tsx
```diff
@@ -49,19 +50,29 @@ const NotificationsPlugin = () => {
...
 
+const DelegatePlugin = () => {
+  return (
+    <div className="col-span-6">
+      <DelegateCard />
+    </div>
+  )
+}
+
 const Account = ({ withHeader = true }: { withHeader?: boolean }) => {
```
SebastianBor, line 61: It's not related to this PR but the `Account` component should not be under `VoterStakeRegistry` namespace

VoteStakeRegistry/components/Account/Account.tsx
```diff
@@ -49,19 +50,29 @@ const NotificationsPlugin = () => {
   return null
 }
 
+const DelegatePlugin = () => {
```
SebastianBor, line 53: nit: the delegation logic is not a plugin

components/DelegateCard.tsx
```diff
@@ -0,0 +1,166 @@
...
+        realm.pubkey, // realm public key
+        tokenMintToDelegateWith, // mint of governance token
+        wallet?.publicKey, // governingTokenOwner (walletId) publicKey of for tokenOwnerRecord of this wallet
+        wallet?.publicKey, // governanceAuthority: publicKey of connected wallet
+        new PublicKey(delegateKey) // public key of wallet who to delegated vote to
+      )
+      const recentBlockhash = await connection.getRecentBlockhash()
+      const transaction = new Transaction({
+        recentBlockhash: recentBlockhash.blockhash,
+      })
```
SebastianBor, line 65: nit: I think the is not needed because `sendTransaction` sets the `blockhash` and `feePayer`

components/DelegateCard.tsx
```diff
@@ -0,0 +1,166 @@
...
+              placeholder="Public key"
+            />
+          </InputRow>
+
+          <Button
+            className="sm:w-full mt-4"
+            onClick={handleDelegate}
+            isLoading={isLoading}
+            disabled={
+              delegateKey.length !== 44 || (!ownTokenRecord && !ownTokenRecord)
```
SebastianBor, line 130: nit: ideally the validation would check if the key exists

components/Members/MemberOverview.tsx
```diff
@@ -160,6 +162,11 @@ const MemberOverview = ({ member }: { member: Member }) => {
   const paginateVotes = (page) => {
     return ownVoteRecords.slice(page * perPage, (page + 1) * perPage)
   }
+
+  const abbreviatePublicKey = (publicKey: string) => {
```
SebastianBor, line 166: nit: don't we have a function for this already? Ideally all address should be displayed using a component with a link to explorer

components/DelegateCard.tsx
ndrabins: Switched to tryParseKey

components/Members/MemberOverview.tsx
ndrabins: removed function

VoteStakeRegistry/components/Account/Account.tsx
ndrabins: removed function

components/DelegateCard.tsx
ndrabins: replaced this with 
```      
const transaction = new Transaction()
transaction.add(...instructions)
```

components/Members/useMembers.tsx
```diff
@@ -210,28 +217,76 @@ export default function useMembers() {
...
+      setDelegates(delegateMap)
       setMembers(members)
     }
-    if (
-      realm?.pubkey &&
-      previousRealmPubKey !== realm?.pubkey.toBase58() &&
-      !realm?.account.config.useCommunityVoterWeightAddin
-    ) {
+    if (realm?.pubkey && previousRealmPubKey !== realm?.pubkey.toBase58()) {
       handleSetMembers()
```
abrzezinski94, line 289: it would be good to separate members and delegates set functions because members are not visible in lock tokens dao yet and its very heavy request so we would not like to call it in locked daos.

components/Members/useMembers.tsx
```diff
@@ -210,28 +217,76 @@ export default function useMembers() {
...
     }
-    if (
-      !realm?.pubkey ||
-      (realm.pubkey && realm?.account.config.useCommunityVoterWeightAddin)
-    ) {
-      setMembers([])
```
abrzezinski94, line 234: if we fix above comment it would be good to leave it for now.

abrzezinski94, line 234: if we fix above comment it would be good to leave it for now.

abrzezinski94 (dismissed):

PR #651, opened 2022-05-03
'Token' on About tab should show mint rather than realm pubkey

emdoyle:
image

The 'Token' section of this tab is actually showing the pubkey of the Realm itself. I believe based on the name it should probably show the community token mint instead.

### DIFF
--- a/components/AboutRealm.tsx
+++ b/components/AboutRealm.tsx
@@ -2,7 +2,7 @@ import React from 'react'
 import useRealm from 'hooks/useRealm'
 
 const AboutRealm = () => {
-  const { realmInfo, realmDisplayName, symbol } = useRealm()
+  const { realmInfo, realmDisplayName, mint } = useRealm()
 
   return (
     <div className="pb-4 space-y-3">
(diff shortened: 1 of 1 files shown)


SebastianBor:
> 
> The 'Token' section of this tab is actually showing the pubkey of the Realm itself. I believe based on the name it should probably show the community token mint instead.

The info is actually correct but the label is wrong. It should be symbol because for the known DAOs it shows it correctly 

image

emdoyle:
> > The 'Token' section of this tab is actually showing the pubkey of the Realm itself. I believe based on the name it should probably show the community token mint instead.
> 
> The info is actually correct but the label is wrong. It should be symbol because for the known DAOs it shows it correctly
> 
> image

Oh, got it! If I were to change the behavior only for uncertified realms/realms that don't provide a token symbol would that be an improvement?

SebastianBor:
> Oh, got it! If I were to change the behavior only for uncertified realms/realms that don't provide a token symbol would that be an improvement?

Let's show the symbol if defined and if not then let's hide it. 
The token info is on the params view

PR #661, opened 2022-05-05
Integration: Castle deposit and withdraw

adamlawrencium:
Summary of changes:
* Added Castle SDK to support `deposit` and `withdraw`
    * additional supporting instructions `refresh` and `reconcile` were also added
* Added `instructionOption` to support the scenario where a user needs execute additional permissionless instructions
* UI button alongside `ExecuteInstructionButton` that allows user to select an `instructionOption`
* Support for executing adjacent transactions (i.e. attempts to send transactions in the same slot)
* Reorganized instruction list so that nonprotocol-specific ones are at the top


utils/instructionTools.ts
```diff
@@ -44,176 +39,6 @@ export const validateInstruction = async ({
   return isValid
 }
 
-export async function getFriktionDepositInstruction({
```
abrzezinski94, line 47: why we want to remove this?

abrzezinski94:
we need to think about some abstraction on exectue instruction ill try to suggest some solution.

utils/instructionTools.ts
adamlawrencium: It's not used anywhere, they moved their instructions to another file. Just doing some clean up in the codebase.

actions/executeTransaction.ts
```diff
@@ -36,15 +56,132 @@ export const executeTransaction = async (
...
+}
+
+/**
+ * Signs and sends proposal transaction with refresh transaction adjacent to it, aiming
+ * for both to be in the same slot.
+ * @param tx Proposal transaction
+ * @param connection
+ * @param wallet
+ */
+const executeWithRefresh = async (
```
SebastianBor, line 98: This function has a generic name `executeWithRefresh()` but it's not a generic function because it's specific to `CastleRefreshInstruction`

Ideally `CastleRefreshInstruction` should be passed in `InstructionOptions` as execution strategy requiring to execute the instruction from a proposal together with other custom instruction and everything in this file `executeTransaction.ts` should remain generic without any Castle specific logic.

For example in the definition of the `Castle Deposit` instruction we could add `Adjacent` instructions, which would be passed here as instruction options
It would work like the `Prerequisite` instructions but they would be executed when the proposal is executed and not created

actions/executeTransaction.ts
```diff
@@ -36,15 +56,132 @@ export const executeTransaction = async (
...
+
+/**
+ * Withdraw requires three transactions
+ * (1) reconcile and refresh ixs
+ * (2) refresh and withdraw ixs.
+ * @param tx Proposal transaction
+ * @param connection
+ * @param wallet
+ */
+const executeWithRefreshAndReconcile = async (
```
SebastianBor, line 136: See the comments for `executeWithRefresh`

actions/executeTransaction.ts
adamlawrencium: Addressed with https://github.com/solana-labs/governance-ui/pull/661/commits/7db7abc6c63b4c27889e503b4d18515be69b0c2e

* Removed all Castle-specific code from the executeTransaction.ts file.
    * added optional parameters for preExecutionTransactions and adjacentTransaction, added documentation
    * added handling these additional parameters
    * moved Castle-specific logic out into a component

PR #718, opened 2022-06-01
Add new DAO wizards 

jonathan3692bf:
Remove old wizard and add three explicit flows for: multi-signature wallets, nft community DAOs, and tokenized DAOs.


components/NewRealmWizard/components/FormField.tsx
```diff
@@ -0,0 +1,198 @@
+// import { useRef, useEffect, useState } from 'react'
+// import { NewButton as Button } from '@components/Button'
+import Text from '@components/Text'
+
+// export function ImageUploader({
```
abrzezinski94, line 5: do we want to leave it commented?

PR #746, opened 2022-06-13
feat: AXY DAO info added to devnet.json

andresmgsl:


andresmgsl:
Thanks @SebastianBor for the quick merge!

PR #775, opened 2022-06-21
Quality of life improvements for lightmode

jonathan3692bf:
- [x] Adjust background images based on theme
- [x] Add new background image for AdviceBox
- [x] Adjust RadioButton disabled state
- [x] Add "warning" when council is forced to be created
- [x] clean up unused assets


asktree:
I am closing this because it's extremely old and I don't think we can prioritize lightmode QoL (it should probably instead just be removed)

PR #802, opened 2022-06-28
Use an indexed rpc endpoint

nramadas:


asktree:
closing because i am going to assume we are no longer going in this direction

Issue #803, opened 2022-06-28
Add Transfer Program Upgrade Authority proposal support


tomjohn1028:
<img width="1523" alt="Screen Shot 2022-06-28 at 5 04 33 PM" src="https://user-images.githubusercontent.com/32071703/176287724-15416883-48f3-455d-b12a-10323a6129e9.png">

Currently Realms does not support transferring program authorities. Transferring program authorities is beneficial if 
A) a DO wants to change current infrastructure to have a separate protocol council realm
B) a DO gets bought out (likely by another DO) and needs to transfer all assets

tomjohn1028:
A DO to DO transfer would require a proposal to create a program governance on the receiving DO. This is because a DO cannot create a program governance without the existing upgrade authority signing the transaction (see [here](https://github.com/solana-labs/solana-program-library/blob/refs/tags/governance-v2.2.5/governance/program/src/processor/process_create_program_governance.rs#L87-L101)).

So the transferrer must do the following
1. `CreateProgramGovernance` instruction for the receiving DO
2.  `SetUpgradeAuthority` instruction for transferrer DO, which takes in the new governance ID from IX 1

tomjohn1028:
To avoid scope creep and move fast, I'm going to limit this issue to a non DO to DO upgrade authority transfer. Will open a separate issue for that after delivering a basic transfer.

PR #805, opened 2022-06-28
Add ability to transfer program authority

tomjohn1028:
closes: https://github.com/solana-labs/governance-ui/issues/803

## Notes

* Even after the authority is transferred the program governance will still show up. Looking at the SPL Governance program there's no close program governance instruction. Because of this, the program still shows up as a governance asset. Is this a PR blocker? Potential solution:
    * Update the governance UI to filter program governances by upgrade authority (ensuring the governance still has the authority

* UI is a little squished on some sizes, but it's actually not terrible (screenshot below). Is this a blocker?
Screen Shot 2022-06-28 at 7 09 10 PM

### DIFF
--- a/components/AssetsList/AssetItem.tsx
+++ b/components/AssetsList/AssetItem.tsx
@@ -14,6 +14,7 @@ import Modal from '@components/Modal'
 import UpgradeProgram from './UpgradeProgram'
 import CloseBuffers from './CloseBuffers'
 import { getExplorerUrl } from '@components/explorer/tools'
+import TransferUpgradeAuthority from './TransferUpgradeAuthority'
 
 const AssetItem = ({
   item,
@@ -26,6 +27,9 @@ const AssetItem = ({
   const [slot, setSlot] = useState(0)
   const [openUpgradeModal, setOpenUpgradeModal] = useState(false)
   const [openCloseBuffersModal, setOpenCloseBuffersModal] = useState(false)
+  const [openTransferAuthorityModal, setOpenTransferAuthorityModal] = useState(
+    false
+  )
   const [loadSlot, setLoadSlot] = useState(false)
   const connection = useWalletStore((s) => s.connection)
   const name = item ? getProgramName(item.account.governedAccount) : ''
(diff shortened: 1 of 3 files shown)


SebastianBor:
>   
>   * Update the governance UI to filter program governances by upgrade authority (ensuring the governance still has the authority

The problem would be with scenarios when somebody creates a program governance but defers authority transfer to some later time. Ideally we should have an option to either hide or close unused governances (not just programs but any others too)

I think for now we could either use the hack with hidden governances https://github.com/solana-labs/governance-ui/blob/0b7f00adaaf9adf9d01196677d6a018b5b22b93f/components/instructions/tools.tsx#L147

Or show the program differently when authority is not transferred. This would be very useful irrespectively of the other options anyway

SebastianBor:
> * UI is a little squished on some sizes, but it's actually not terrible (screenshot below). Is this a blocker?
> 
> Screen Shot 2022-06-28 at 7 09 10 PM

The logic looks ok, for the UI it's not a blocker and can be merged for now but we have to do some design work on our side for scenarios where multiple context actions are used

tomjohn1028:
>> Update the governance UI to filter program governances by upgrade authority (ensuring the governance still has the authority
>
> The problem would be with scenarios when somebody creates a program governance but defers authority transfer to some later time. Ideally we should have an option to either hide or close unused governances (not just programs but any others too)
>
> I think for now we could either use the hack with hidden governances

@SebastianBor How about the following? It hides the action items and displays the following message when the upgrade authority does not match the governance. This implementation also doesn't require any additional network requests! ([commit](https://github.com/solana-labs/governance-ui/pull/805/commits/5aeb8a1914da34eb96e1fb6af792615c8b77be9e))

Screen Shot 2022-06-28 at 9 09 10 PM

SebastianBor:
> 
> @SebastianBor How about the following? It hides the action items and displays the following message when the upgrade authority does not match the governance. This implementation also doesn't require any additional network requests! ([commit](https://github.com/solana-labs/governance-ui/pull/805/commits/5aeb8a1914da34eb96e1fb6af792615c8b77be9e))
> 

That works

tomjohn1028:
This is good to go then

PR #813, opened 2022-06-30
upgrading Profile pic

DimiMili:
upgrading Profile pic


SebastianBor (commented):
It looks like there are 2 extra unnecessary icons. Do we need 3 of them?

DimiMili:
No we do not need 3 of them. We just need one. I mistakenly uploaded more than once.

SebastianBor:
> No we do not need 3 of them. We just need one. I mistakenly uploaded more than once.

ok, then please remove the extra pictures from the PR or create a new one with just the picture which is needed

DimiMili:
I think I deleted them. (first time i'm doing sth like this). Let me know if it's ok now.

SebastianBor:
> I think I deleted them. (first time i'm doing sth like this). Let me know if it's ok now.

Almost there :) There is still one extra public/GrapeCCSubDAO/GrapeCCSubDAO.png

PR #828, opened 2022-07-07
Add Create Token Metadata and Update Token Metadata

KishiTheMechanic:
Added Create Token Metadata and Update Token Metadata with mpl-token-metadata v2.
We can update the metadata of community token and council token of the DAO on the governance-ui.

PR #866, opened 2022-07-16
Create / Update Token Metadata

KishiTheMechanic:
From #828


KishiTheMechanic:
Fixed about making SOL treasury logic and selecting correct governance account when executing.

PR #889, opened 2022-07-25
Display spl token account owner wallet

nemanjab17:
Hi, we have had multiple requests from DAO owners to create streamflow contracts and stream spl tokens directly to treasury token accounts. On our UI, we use simple input field for receiver's wallet, then we derrive ATA and use it as token account. Currently, users cannot find their treasury's owner account. This change renders owner.

### DIFF
--- a/components/TreasuryAccount/AccountOverview.tsx
+++ b/components/TreasuryAccount/AccountOverview.tsx
@@ -412,6 +412,12 @@ const AccountOverview = () => {
           </a>
         </div>
       </div>
+      {isSplToken && (
+        <p className={`pb-4`}>
+          Owner:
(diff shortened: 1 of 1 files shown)


SebastianBor:
This won't solve the problem because those token accounts are not necessarily ATAs and it would create even more confusion.  
We are currently working on a new version of the screen to show the relationship and it would address this problem.
In the meantime if you wanted to help your users then instead of displaying the owner account, show the owner DAO wallet (the SOL account) and only if the given token account is in fact ATA
You can add ATA to any DAO wallet using the 'Add new token account' button 
image

nemanjab17:
Okay, let's wait for the new solution and see how it looks like, closing this one

PR #893, opened 2022-07-27
addition of Logo for Dean's List Realm

DimiMili:


SebastianBor:
Which logo do you want to add/change? There are two png files but looks like only `GrapeCCSubDAO.png` is used

DimiMili:
Grape CC logo is ok. Want to add the logo to Dean's List realm.

PR #921, opened 2022-08-03
changed back to show address instead of token name on treasury tokens

### LINKED ISSUE #920: Can't show Token Metadata correctly on mainnet
image
image

It seems different mainnet and dev.
Need to check what is happening.

KishiTheMechanic:
changed back to show address instead of token name on treasury tokens

Also add Epics on mainnet-beta.json
and
Hide "Epics DAO" because we lost voting power for that.


KishiTheMechanic:
Changed the way to get token metadata image URI.

Before I used deprecated function for this but now I changed it to latest version.
In dev works as fine.
So let's see if it fix #920

PR #923, opened 2022-08-03
update Epics DAO and fix to show token metadata image on treasury

KishiTheMechanic:
removed banner.
changed to use svg.

About token metadata image, still talking with metaplex team.
They will fix it or asking another way to get token metadata now.

Thank you always!
Have a good day.


KishiTheMechanic:
Finally we can Close #920

PR #930, opened 2022-08-09
use text input for create ATA mint selection

tomjohn1028:
image

* the current Select drop down is too limiting


SebastianBor:
> 
> * the current Select drop down is too limiting

Could you fetch the token and show what it resolves to (same as in the 'Create token account' view) please?

image

tomjohn1028:
>Could you fetch the token and show what it resolves to (same as in the 'Create token account' view) please?


I'm going to refactor to make that Token Mint input a reusable component. Closing this PR for now.

tomjohn1028:
@SebastianBor much improved with reusable `TokenMintInput` component

Create Token
Screen Shot 2022-08-09 at 4 12 27 PM

Create ATA proposal
Screen Shot 2022-08-09 at 4 12 50 PM
l

PR #1033, opened 2022-09-12
DAO asset config icons

asktree:
Here I update the DAO asset config UI a bit to bring it up to par with the Wallet Rules UI.

Before:
image

After: 
image

For reference, here is Wallet Rules:
image

I also made the asset UI use the `Section` component for typography, from the Wallet UI. Usually I am not such a fan of recycling components, but I think in this case we would always want to maintain identical look and feel. You'll notice some subtle differences in this typography from the original asset config, such as the font weight of the settings.


components/treasuryV2/Details/RealmAuthorityDetails/Config.tsx
```diff
@@ -23,46 +30,53 @@ export default function Config(props: Props) {
...
+            title: 'Community mint max vote weight source',
+            value: props.realmAuthority.config.communityMintMaxVoteWeightSource.fmtSupplyFractionPercentage(),
+            icon: <ScaleIcon />,
+          },
+        ]
+      : []),
+    {
+      icon: <DocumentAddIcon />,
+      title: 'Min community tokens to create governance',
+      value: DISABLED.shiftedBy(
```
SebastianBor, line 46: It's not related to this change but wanted to highlight the issue in this code.
The logic is wrong because the option is only disabled when `minCommunityTokensToCreateGovernance` == `DISABLED`

components/treasuryV2/Details/RealmAuthorityDetails/Config.tsx
```diff
@@ -23,46 +30,53 @@ export default function Config(props: Props) {
...
+        -(mint ? mint.decimals : 0)
+      ).isLessThanOrEqualTo(
+        props.realmAuthority.config.minCommunityTokensToCreateGovernance
+      )
+        ? 'Disabled'
+        : formatNumber(
```
SebastianBor, line 52: Same here, this should be addressed in a follow up PR
The format should show the value using mint decimals and not as is

SebastianBor (approved):
Looks good, I only flagged  existing issues which should be addressed in a follow up PR

PR #1035, opened 2022-09-12
Agrippa/no-ugly-numbers

asktree:
I have created a new input type for FormCreator.tsx, `DISABLEABLE_INPUT`. 

This essentially copy-pastes the UI from the BaseGovernanceForm for interacting with disabled values, though the formatting is slightly different and keeps things on one line. 

image

(black text unrelated)


asktree:
Here are 2 test proposals I created with my UI:

✅ Disabling:
http://localhost:3000/dao/12aXtp1prbu7AXfzPwyVQHsPyqkqCwovjMRZufiHoocU/proposal/HH5g8uhHaiFbC7U9PkG8XXLiY8pTponmfMgQEWFmkqm8?cluster=devnet

✅ Setting value to 3 tokens:
http://localhost:3000/dao/12aXtp1prbu7AXfzPwyVQHsPyqkqCwovjMRZufiHoocU/proposal/B3S9sWDZUZbCrPevWf8QmFmgD4APkShXN4qeHQuByzPC?cluster=devnet

tools/constants.ts
```diff
@@ -11,3 +11,4 @@ export const MIN_COMMUNITY_TOKENS_TO_CREATE_W_0_SUPPLY = 1000000
  */
 
 export const DISABLED_VOTER_WEIGHT = new BN('18446744073709551615')
+export const DISABLED_VALUE = new BN('18446744073709551615')
```
asktree, line 14: I felt this was more accurate, since this value is used for more than just disabled weights. We should probably not have two names for the same constant though.

SebastianBor: It's actually used for disabled voter weights only. It denotes the value (`u64::MAX`) which is used to set given option as disabled. You can find more details here https://github.com/solana-labs/solana-program-library/pull/3501

In the codebase you might find legacy names `minTokensToCreateProposal`, `minTokensToCreateGovernance`. They are legacy because the concept of the thresholds and governance power was abstracted away from the number of tokens to abstract weight which might or might not be related to tokens

SebastianBor (approved):

PR #1046, opened 2022-09-14
add skip script from vercel examples

asktree:
This is for usage in blockworks vercel config, for skipping preview deploys.


asktree:
image
here you can see the output from my script in the deply log when I merged to main, thus confirming that production deploys still work.

asktree:
Verifying that previews dont build any longer: https://github.com/solana-labs/governance-ui/pull/1047

PR #1112, opened 2022-10-24
Add Squads to about.json

emoney023:


public/realms/about.json
```diff
@@ -178,5 +178,21 @@
         "answer": ["Yes we have three."]
       }
     ]
+  },
+    "BzGL6wbCvBisQ7s1cNQvDGZwDRWwKK6bhrV93RYdetzJ": {
+    "symbol": "Squads",
```
nramadas, line 183: you have a syntax error with this one. lines 183 - 196 should be indented. If you do that, you'll notice that you're missing a closing `}` between lines 196 & 197

PR #1141, opened 2022-10-27
Realms is now compatible with mobile wallet apps

nramadas:


abrzezinski94:
all moved to https://github.com/solana-labs/governance-ui/pull/1145

PR #1189, opened 2022-11-04
use map instead of switch/case for instruction/component mapping

GregoryNEUT:
This PR aim to change how we map instructions ids with theirs components.

Today it is a switch/case, I propose to use a strong typed map.

Benefits: If someone declare a new instruction in enum Instructions then it shows automatically an error in the pages/dao/[symbol]/proposal/new.tsx file. Hope it helps developper implementing new instructions.

Demo here:

https://user-images.githubusercontent.com/22965416/199916700-b4b6eef3-3cf8-4a01-99ce-28d3f76148dd.mov


abrzezinski94:
looks good i like it

PR #1208, opened 2022-11-06
Introduce react-query + use it to discriminate program version in the creation wizard

asktree:
# about `react-query`
I've been looking to introduce `react-query` into the codebase because it's super useful. I ran into a case (fetching and caching the programVersion of a Realms program) that wasn't covered by existing query stores, and thought it would be a good chance to do so.

By using `react-query` we can manage queries and caches on a pubkey-by-pubkey basis without needing to maintain a monolithic data store. `react-query` will also give us a really easy API for expressing how long queries should be cached or made stale.

# feature
The "Custom Program ID" input now has a message displaying the version of the input program
image

Here's what happens when I put in a random pubkey:
image

(there is also a loading message while fetching, but I was not fast enough to capture it).

## implementation notes
This is implemented using a `useProgramVersionByIdQuery` hook. While results are only cached for 5 minutes, they never expire, and this hook can be reused anywhere else in the app where we would want to get programVersion given a program id. 

I intend to use this same hook in order to conditionally display form steps based on the program version.

### DIFF
--- a/components/NewRealmWizard/components/steps/BasicDetailsForm.tsx
+++ b/components/NewRealmWizard/components/steps/BasicDetailsForm.tsx
@@ -60,7 +60,7 @@ export default function BasicDetailsForm({
     setValue,
     control,
     handleSubmit,
-    formState: { errors, isValid, touchedFields },
+    formState: { errors, isValid },
   } = useForm({
     mode: 'all',
     resolver: yupResolver(schema),
--- a/components/NewRealmWizard/components/steps/BasicDetailsForm.tsx
+++ b/components/NewRealmWizard/components/steps/BasicDetailsForm.tsx
@@ -136,6 +136,20 @@ export default function BasicDetailsForm({
                   data-testid="programId-input"
                   error={errors.programId?.message || ''}
                   autoComplete="on"
+                  success={
+                    !programVersionQuery.isLoading &&
+                    programVersionQuery.data !== 1
+                      ? `Program version ${programVersionQuery.data}`
+                      : undefined
+                  }
+                  warning={
+                    !programVersionQuery.isLoading &&
+                    programVersionQuery.data === 1
+                      ? 'Program version could not be verified'
+                      : programVersionQuery.isFetching
+                      ? 'Fetching program version...'
+                      : undefined
+                  }
                   {...field}
                 />
               </FormField>
--- a/components/NewRealmWizard/components/steps/BasicDetailsForm.tsx
+++ b/components/NewRealmWizard/components/steps/BasicDetailsForm.tsx
@@ -1,5 +1,5 @@
 import { useEffect } from 'react'
-import { useForm, Controller } from 'react-hook-form'
+import { useForm, Controller, useWatch } from 'react-hook-form'
 import { yupResolver } from '@hookform/resolvers/yup'
 import * as yup from 'yup'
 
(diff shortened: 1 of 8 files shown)


hooks/queries/queryClient.ts
```diff
@@ -0,0 +1,3 @@
+import { QueryClient } from '@tanstack/react-query'
+
+export default new QueryClient()
```
asktree, line 3: It's often useful to be able to directly access the global `queryClient`, so I just gave it its own file. Alternatively people can use `useQueryClient` and this could've been a private variable in the provider file.

hooks/queries/useProgramVersionQuery.ts
```diff
@@ -0,0 +1,29 @@
+import { getGovernanceProgramVersion } from '@solana/spl-governance'
+import { PublicKey } from '@solana/web3.js'
+import { useQuery } from '@tanstack/react-query'
+import useWalletStore from 'stores/useWalletStore'
+
+export const programVersionQueryKeys = {
```
asktree, line 6: in `react-query` we manage caching using `queryKey`s. 
See https://tanstack.com/query/v4/docs/guides/query-keys

I personally like to use this factory pattern for `queryKey`s in order to eliminate surface area for typos, but the usefulness of this is not so evident here given that my `queryKey` only gets used once.

hooks/queries/useProgramVersionQuery.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export function useProgramVersionByIdQuery(realmsProgramId?: PublicKey) {
+  // @asktree is unsure why we use this instead of `useConnection` (which has no corresponding provider in the BaseApp)
+  const { connection } = useWalletStore()
+  console.log('connection', connection)
+
+  const query = useQuery({
+    queryKey:
+      realmsProgramId && programVersionQueryKeys.byProgramId(realmsProgramId),
+    queryFn: () =>
```
asktree, line 21: the query itself is quite simple, since I'm just using an already-existing function `getGovernanceProgramVersion`. By essentially just wrapping it in a `useQuery` we get intelligent and easy caching, an easy api for knowing query status, etcetera.

pages/_app.tsx
```diff
@@ -23,22 +25,19 @@ export default function App({ Component, pageProps, router }: AppProps) {
...
-        <Component {...pageProps} />
-      </HubApp>
-    )
-  }
-
   return (
-    <BaseApp>
-      <Component {...pageProps} />
-    </BaseApp>
+    <QueryClientProvider client={queryClient}>
```
asktree, line 29: As far as I can tell there is no reason that `HubApp` and `BaseApp` should not share a `QueryClientProvider`

components/NewRealmWizard/components/steps/BasicDetailsForm.tsx
```diff
@@ -116,6 +136,20 @@ export default function BasicDetailsForm({
...
+                    !programVersionQuery.isLoading &&
+                    programVersionQuery.data !== 1
+                      ? `Program version ${programVersionQuery.data}`
+                      : undefined
+                  }
+                  warning={
+                    !programVersionQuery.isLoading &&
+                    programVersionQuery.data === 1
+                      ? 'Program version could not be verified'
+                      : programVersionQuery.isFetching
```
asktree, line 149: The difference between `isLoading` and `isFetching` is that `isLoading` is true iff we have no cached data, and so will be true even if no fetching is taking place. `react-query` gives us an answer to every imaginable question we might ask about the state of this query.

hooks/queries/useProgramVersionQuery.ts
```diff
@@ -0,0 +1,29 @@
...
+      getGovernanceProgramVersion(connection.current, realmsProgramId!),
+    enabled: realmsProgramId !== undefined,
```
asktree, line 23: We can stop the queryFn from executing if the input is undefined, thus letting us make optional queries.

asktree:
I fixed the linter issue that Vercel was complaining about in previous checks and am not sure how to make it re-check.

abrzezinski94:
Nice work, looks good to me we need to wait for @SebastianBor to look

PR #1218, opened 2022-11-13
use indexer to get realms

abrzezinski94:


models/registry/api.ts
```diff
@@ -173,10 +174,14 @@ export async function getUnchartedRealmInfos(connection: ConnectionContext) {
   ).map((p) => {
     return p.programId
   })
-
+  const programIdsChunk = chunks(programIds, 5)
```
GregoryNEUT, line 177: note: `chunks` from `@utils/helpers`  and `group` and `@utils/group` seems to do the same thing.

abrzezinski94: Hi actually chunk is not helping so we will use indexer waiting for them to update query for us :)

PR #1225, opened 2022-11-16
Feature/metaplex upgrade

abrzezinski94:


components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
```diff
@@ -59,11 +71,16 @@ async function enrichItemInfo(item, uri) {
   }
 }
 
-async function enrichCollectionInfo(connection, collectionKey) {
-  const data = await deprecated.Metadata.findByMint(connection, collectionKey)
-
-  const collectionData = data!.data!.data
+async function enrichCollectionInfo(
+  connection: Connection,
+  collectionKey: string
```
GregoryNEUT, line 76: May be interesting to use `collectionKey: PublicKey` here.

We use `toBase58()` on PublicKey to then re-transform the string to PublicKey

The actual call:
```
const collectionInfo = await enrichCollectionInfo(
      connection,
      metaplexData.collection.address.toBase58()
    )
```

components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
```diff
@@ -73,36 +90,64 @@ async function enrichCollectionInfo(connection, collectionKey) {
...
+    },
+    body: JSON.stringify({
+      query: getNftQuery(1, 0),
+      variables: {
+        collection: new PublicKey(collectionKey),
+      },
+    }),
+  })
+  const body = await result.json()
+  const mintToCheck = body.data?.nfts[0]?.mintAddress
```
GregoryNEUT, line 110: `mintToCheck` may be undefined here. I think `new PublicKey(mintToCheck)` would fail.

Maybe throw a custom error to ease debugging if it happens? Or handle having no data if that's an accepted case

components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
```diff
@@ -73,36 +90,64 @@ async function enrichCollectionInfo(connection, collectionKey) {
...
     )
-    const nft = await enrichItemInfo(result.data, result.data.uri)
+    const nft = await enrichItemInfo(metaplexData, metaplexData.uri)
     collectionInfo.nfts = [nft]
     return collectionInfo
   } else {
     // assume we've been given the collection address already, so we need to go find it's children
-    const children = await deprecated.Metadata.findMany(connection, {
-      updateAuthority: result!.updateAuthority,
+    const result = await fetch('https://graph.holaplex.com/v1', {
```
GregoryNEUT, line 128: may use `axios.post` it reduces the noise around `method`, `headers` etc.

I think something like:

```
const result = await axios.post(
    'https://graph.holaplex.com/v1',
    {
        query: getNftQuery(1000, 0),
        variables: {
          collection: new PublicKey(collectionKey),
        },
     },
  )
```

GregoryNEUT (commented):
Hello, a small review, hope it helps! 🙏

components/TreasuryAccount/DepositNFTAddress.tsx
```diff
@@ -99,11 +105,11 @@ const DepositNFTAddress = ({ additionalBtns }: { additionalBtns?: any }) => {
...
           try {
-            const metadataPDA = await deprecated.Metadata.getPDA(pubKey)
-            const tokenMetadata = await deprecated.Metadata.load(
-              connection.current,
-              metadataPDA
-            )
+            const metaplex = new Metaplex(connection.current)
+            const metadataPDA = findMetadataPda(pubKey)
+            const tokenMetadata = await metaplex.nfts().findByMetadata({
+              metadata: new PublicKey(metadataPDA.toBase58()),
```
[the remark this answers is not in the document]
abrzezinski94: Yes metadataPda is PDA class with bump so its diffrent from PublicKey there is option to do toBase58() on that or cast it as PublicKey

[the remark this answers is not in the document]
abrzezinski94: Ok new linter is not throwing error on that gonna check it

components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
abrzezinski94: This is one thing that i will left because in whole component it is mostly used as string for view or operations and whole component has most of things as any... so when i change it to pubkey it will break things that im not aware of.  It would be good to rewrite a bit the logic in there and add normal typing becuase in some places it seems weird with new metaplex but i don't want to do it now.

PR #1230, opened 2022-11-18
[security] Freeze packages

asktree:
# Motivation
NPM packages are a vector for crypto heists. This makes sense because by default tools like `yarn` and `npm` frequently update packages according to semantic versioning logic. This opens us up to a few attacks:
1) a malicious contribution is made to the repo which adds a non-malicious "sleeper agent" package, that is later updated with malicious code. We're especially susceptible to this since we let DAOs add their own packages.
2) a non-malicious "sleeper agent" package is made and adopted by the ecosystem because it's useful and later updated with malicious code.
3) a useful package gets adoption and then handed off to a malicious volunteer, who updates it later with malicious code. This happened semi-recently on Solana or Eth, couldn't find an article about it.

# Solutions
### Always install exact package versions
I have changed package.json to only contain exact versions. This is probably overkill for major packages like `react`, but I wanted the norm to be really clear for now. I also added `.yarnrc` config that makes exact versioning the default when using `yarn add`. 

### Always use `yarn install` with frozen lockfile
For some reason, yarn by default can update dependencies in the lockfile when using `yarn install`. I think this is changed in future / cutting-edge versions of yarn. I've changed `.yarnrc` to make this our default. 

# Caveat
Installing exact versions of packages is not enough for a total freeze on dependencies, because *those* packages can still have non-exact dependencies which yarn can update silently. Our build system will not do this, but it could happen in the course of development. As far as I can tell there is no canonical solution for this using `yarn`, but I think `npm` with `npm shrinkwrap` does a deep freeze. We may in the future decide this is the best path, further research needed.

The above should make us a lot more resilient against malicious updates. I think contributors and reviewers should make a habit of inspecting yarn.lock for semantic updates of packages, before merging.


.yarnrc.yml
```diff
@@ -1 +0,0 @@
-checksumBehavior: 'update'
```
asktree, line 1: Yarn.lock contains an integrity checksum for each package, so that if the package is updated without a version change it will fail to install.

`checksumBehavior` dictates what happens when an installed package fails the integrity check. We want the integrity check to throw errors! Updating the checksum to match the new one is the opposite of what we want.

Strangely you cannot use `.yarnrc.yml` for CLI arguments like you can `.yarnrc` so I just deleted the file entirely.

yarn.lock
```diff
@@ -1748,12 +1761,12 @@
   resolved "https://registry.yarnpkg.com/@headlessui/react/-/react-1.7.3.tgz#853c598ff47b37cdd192c5cbee890d9b610c3ec0"
   integrity sha512-LGp06SrGv7BMaIQlTs8s2G06moqkI0cb0b8stgq7KZ3xcHdH3qMP+cRyV7qe5x4XEW/IGY48BW4fLesD6NQLng==
 
-"@heroicons/react@^1.0.1":
-  version "1.0.6"
-  resolved "https://registry.yarnpkg.com/@heroicons/react/-/react-1.0.6.tgz#35dd26987228b39ef2316db3b1245c42eb19e324"
-  integrity sha512-JJCXydOFWMDpCP4q13iEplA503MQO3xLoZiKum+955ZCtHINWnx26CUxVxxFQu/uLb4LW3ge15ZpzIkXKkJ8oQ==
+"@heroicons/react@1.0.1":
+  version "1.0.1"
```
abrzezinski94, line 1765: probably 1.0.6 ?

yarn.lock
```diff
@@ -2567,91 +2621,91 @@
...
-  resolved "https://registry.yarnpkg.com/@next/swc-win32-ia32-msvc/-/swc-win32-ia32-msvc-12.3.2.tgz#769bef60d0d678c3d7606a4dc7fee018d6199227"
-  integrity sha512-5hRUSvn3MdQ4nVRu1rmKxq5YJzpTtZfaC/NyGw6wa4NSF1noUn/pdQGUr+I5Qz3CZkd1gZzzC0eaXQHlrk0E2g==
-
-"@next/swc-win32-x64-msvc@12.3.2":
-  version "12.3.2"
-  resolved "https://registry.yarnpkg.com/@next/swc-win32-x64-msvc/-/swc-win32-x64-msvc-12.3.2.tgz#45beb4b9d28e6dd6abf63cab1c5b92dc84323a6b"
-  integrity sha512-tpQJYUH+TzPMIsdVl9fH8uDg47iwiNjKY+8e9da3dXqlkztKzjSw0OwSADoqh3KrifplXeKSta+BBGLdBqg3sg==
-
-"@nfteyez/sol-rayz@^0.10.2":
+"@next/env@12.2.6":
```
abrzezinski94, line 2636: at least 12.3.2

yarn.lock
```diff
@@ -3353,21 +3407,21 @@
     bn.js "^5.1.2"
     buffer-layout "^1.2.0"
 
-"@project-serum/serum@0.13.64":
-  version "0.13.64"
-  resolved "https://registry.yarnpkg.com/@project-serum/serum/-/serum-0.13.64.tgz#d5a9009ded8d2539ebfd12e1ff58b9028cfcdf45"
-  integrity sha512-33d8K3qcbiFnL1Azv5k6IYMtJYjCqqmFkf95w5W8MPB0A+5zQfOwhEfXQuBw0ExZ8ft5s9Vy8sReLyx0SsINWA==
+"@project-serum/serum@0.13.61", "@project-serum/serum@^0.13.21", "@project-serum/serum@^0.13.61":
```
abrzezinski94, line 3410: at least 0.13.64

yarn.lock
```diff
@@ -5824,10 +5885,10 @@
   resolved "https://registry.yarnpkg.com/@supercharge/promise-pool/-/promise-pool-2.3.0.tgz#9528b11d69a63ff2a342d404afb13cd011c9651a"
   integrity sha512-KLf3JBWW4JymplvT/c5JCZhc4KrNE9EBf15FPjvklQiGcoUczSLInpV23iV5dITnvktJlwQ5xq08wfNwBk71Rg==
 
-"@swc/helpers@0.4.11":
-  version "0.4.11"
-  resolved "https://registry.yarnpkg.com/@swc/helpers/-/helpers-0.4.11.tgz#db23a376761b3d31c26502122f349a21b592c8de"
-  integrity sha512-rEUrBSGIoSFuYxwBYtlUFMlE2CwGhmW+w9355/5oduSw8e5h2+Tj4UrAGNNgP9915++wj5vkQo0UuOBqOAq4nw==
+"@swc/helpers@0.4.3":
```
abrzezinski94, line 5888: at least 0.4.11

yarn.lock
```diff
@@ -10524,12 +10631,12 @@ eslint-visitor-keys@^3.3.0:
   resolved "https://registry.yarnpkg.com/eslint-visitor-keys/-/eslint-visitor-keys-3.3.0.tgz#f6480fa6b1f30efe2d1968aa8ac745b862469826"
   integrity sha512-mQ+suqKJVyeuwGYHAdjMFqjCyfl8+Ldnxuyp3ldiMBFKkvytrXUZWaiPCEav8qDHKty44bD+qV1IP4T+w+xXRA==
 
-eslint@^8.14.0:
-  version "8.17.0"
-  resolved "https://registry.yarnpkg.com/eslint/-/eslint-8.17.0.tgz#1cfc4b6b6912f77d24b874ca1506b0fe09328c21"
-  integrity sha512-gq0m0BTJfci60Fz4nczYxNAlED+sMcihltndR8t9t1evnU/azx53x3t2UHXC/uRjcbvRw/XctpaNygSTcQD+Iw==
+eslint@8.14.0:
+  version "8.14.0"
```
abrzezinski94, line 10635: at least 8.17.0

yarn.lock
```diff
@@ -14905,49 +15175,49 @@ neo-async@^2.6.0:
...
   dependencies:
     enhanced-resolve "^5.10.0"
     escalade "^3.1.1"
 
-next@^12.2.6:
-  version "12.3.2"
-  resolved "https://registry.yarnpkg.com/next/-/next-12.3.2.tgz#3a3356a8d752726128825a8bdf17f2a3b3f861cf"
-  integrity sha512-orzvvebCwOqaz1eA5ZA0R5dbKxqtJyw7yeig7kDspu6p8OrplfyelzpvMHcDTKscv/l0nn/0l0v3mSsE8w4k7A==
+next@12.2.6:
+  version "12.2.6"
```
abrzezinski94, line 15197: at least 12.3.2?

yarn.lock
```diff
@@ -1154,14 +1167,14 @@
   version "0.2.5"
   resolved "https://registry.npmjs.org/@emotion/weak-memoize/-/weak-memoize-0.2.5.tgz"
 
-"@eslint/eslintrc@^1.3.0":
-  version "1.3.0"
-  resolved "https://registry.yarnpkg.com/@eslint/eslintrc/-/eslintrc-1.3.0.tgz#29f92c30bb3e771e4a2048c95fa6855392dfac4f"
-  integrity sha512-UWW0TMTmk2d7hLcWD1/e2g5HDM/HQ3csaLSqXCfqwh4uNDuNqlaKWXmEsL4Cs41Z0KnILNvwbHAah3C2yt06kw==
+"@eslint/eslintrc@^1.2.2":
```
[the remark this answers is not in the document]
asktree: This package isn't in our package.json

yarn.lock
asktree: good catch

PR #1249, opened 2022-11-25
use and add yarn-deduplicate

asktree:
Includes a note on postinstall to consider running `yarn-deduplicate`

(There is no simple way to make it automatically run only after `yarn add`)

image

About: https://medium.com/@bnaya/yarn-deduplicate-the-hero-we-need-f4497a362128
If you want to see the difference it makes, just check out the changes to yarn.lock!


abrzezinski94:
made this fix part of https://github.com/solana-labs/governance-ui/pull/1248

PR #1256, opened 2022-11-28
[stacked pr] Veto UI

asktree:
I rebased and made a new PR https://github.com/solana-labs/governance-ui/pull/1295
---
Here is a list of feature tickets this PR implements:
- Allow proposals to be filtered by "Vetoed"
- Show veto progress for veto-able proposals
- Show your veto votes when viewing a proposal
- Allow users to veto
- Show vetoed status throughout the app for vetoed proposals

I also implemented some migration to `react-query` that should grant quite large improvements to fetching user VoteRecords.

It does not implement: 
- Making the proposal explorer show veto votes in graphs etc.
- Vetoing proposals when using batch voting
- Showing any detailed information about veto status outside of the proposal page

# Screenshots
From http://localhost:3000/dao/6NAdM6hYDr8rqBYwXiaPFxyEbv3T81nVQ57tvQgdGh1w?cluster=devnet
image
image
image
image

# Testing
Try (I also did all of the below): 
- vetoing a proposal 
- voting normally
- v2 regression tests

Also check out places that might break due to the existence of a new vote type and proposal state. It is very hard to check for all of them because I don't know where the are 😦. In the future maybe it is good to use exhaustiveness checks that typescript can read, I implemented this in cases that I did spot (see assertUnreachable.ts).


components/ProposalStateBadge.tsx
```diff
@@ -46,6 +48,8 @@ function getBorderColor(proposalState: ProposalState, otherState: OtherState) {
         : 'border-[#5DC9EB]'
     case ProposalState.Voting:
       return otherState.votingEnded ? 'border-[#5DC9EB]' : 'border-[#8EFFDD]'
+    default:
```
asktree, line 51: if proposalStates ever change again in the future then this will error and remind us to be exhaustive.

components/treasuryV2/Details/TokenOwnerRecordDetails/ProposalDetails.tsx
```diff
@@ -26,104 +26,12 @@ import {
 import { PublicKey, Transaction, TransactionInstruction } from '@solana/web3.js'
 import { notify } from '@utils/notifications'
 import { InstructionDataWithHoldUpTime } from 'actions/createProposal'
-import classNames from 'classnames'
 import Link from 'next/link'
 import { useRouter } from 'next/router'
 import { useCallback, useState } from 'react'
 import useWalletStore from 'stores/useWalletStore'
 import VoteProposalModal from './VoteProposalModal'
-
```
asktree, line 35: this stuff was identical to a component in another file

hooks/useProposalVotes.tsx
```diff
@@ -18,23 +24,31 @@ export default function useProposalVotes(proposal?: Proposal) {
   // TODO: optimize using memo
   if (!realm || !proposal || !governance || !proposalMint)
     return {
-      voteThresholdPct: 100,
```
asktree, line 21: in my opinion hooks should not obfuscate whether data is loading or just an optimistic default, so here i made the components individually decide how to handle loading data.

pages/dao/[symbol]/index.tsx
```diff
@@ -266,14 +266,6 @@ const REALM = () => {
     setIsMultiVoting(false)
   }
 
-  const showMultiVote = useMemo(
```
Comment (asktree) [inline pages/dao/[symbol]/index.tsx:269]: minor change after chatting sebastian -- batch voting should just be visible always. its not like gmail disables the batch operation UI just because you only have 1 email in the inbox.

components/VotePanel/hooks.ts
```diff
@@ -0,0 +1,66 @@
...
+  const { ownTokenRecord, ownCouncilTokenRecord } = useRealm()
+
+  const voterTokenRecord =
+    tokenRole === GoverningTokenRole.Community
+      ? ownTokenRecord
+      : ownCouncilTokenRecord
+  return voterTokenRecord
+}
+
+export const useProposalVoteRecordQuery = (quorum: 'electoral' | 'veto') => {
```
asktree, line 41: Here I have rewritten our code for fetching user `VoteRecord`s using react-query. It is:
- self-contained
- faster than previous (useWalletStore actually uses a getProgramAccounts RPC call and then filters it down to one account for some reason which is way way slower)
- much more complete API 

I do not migrate the entire codebase to use this new hook.

hooks/queries/addresses/voteRecord.ts
```diff
@@ -0,0 +1,54 @@
+import useRealm from '@hooks/useRealm'
+import { getVoteRecordAddress } from '@solana/spl-governance'
+import { PublicKey } from '@solana/web3.js'
+import { useQuery } from '@tanstack/react-query'
+import { useRouter } from 'next/router'
+
+export const useAddressQuery_SelectedProposalVoteRecord = (
```
asktree, line 7: here is an example of composing queries

.husky/pre-commit
```diff
@@ -1,4 +1,2 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
-
-yarn lint-staged
```
asktree, line 4: I do kill husky in this PR, can undo if you want @SebastianBor.

PR #1270, opened 2022-12-01
Fix proposal warnings for daos using SPL v1

nramadas:


pages/dao/[symbol]/proposal/[pk]/ProposalWarnings.tsx
```diff
@@ -62,7 +62,7 @@ const useProposalSafetyCheck = () => {
     if (realmInfo === undefined || config === undefined) return undefined
 
     const ixs = Object.values(instructions).flatMap(
-      (pix) => pix.account.instructions
+      (pix) => pix.account.instructions || [pix.account.instruction]
```
Comment (SebastianBor) [inline pages/dao/[symbol]/proposal/[pk]/ProposalWarnings.tsx:65]: it would be better to use getAllInstructions() instead of accessing individual fields

PR #1279, opened 2022-12-07
Integrate Jupiter Terminal Modal

dicksonpys:


hub/components/RealmHeader/index.tsx
```diff
@@ -11,13 +11,17 @@ import OverflowMenuHorizontalIcon from '@carbon/icons-react/lib/OverflowMenuHori
...
 import * as NavigationMenu from '@radix-ui/react-navigation-menu';
+import { useWallet } from '@solana/wallet-adapter-react';
 import type { PublicKey } from '@solana/web3.js';
 import { pipe } from 'fp-ts/lib/function';
 import Link from 'next/link';
-import { useMemo } from 'react';
+import { useCallback, useMemo } from 'react';
 import { useMediaQuery } from 'react-responsive';
 
+import useWalletStore from 'stores/useWalletStore';
```
nramadas, line 23: to get the `connection`, please use `https://github.com/solana-labs/governance-ui/blob/main/hub/hooks/useCluster.ts` instead:

```ts
import { useCluster } from '@hub/hooks/useCluster';

const [cluster] = useCluster();
const connection = cluster.connection;
```

pages/_document.tsx
```diff
@@ -66,6 +66,12 @@ class RealmsDocument extends Document {
           <meta name="twitter:title" content="Realms" />
           <meta property="og:type" content="website" />
           <meta name="twitter:card" content="summary" />
+
+          <script
```
nramadas, line 70: can we import these dynamically instead when the user clicks on the button?

hub/components/RealmHeader/index.tsx
```diff
@@ -60,12 +64,31 @@ interface Props extends BaseProps {
...
 export function Content(props: Props) {
-  const jupiterDirectLink = useMemo(() => {
-    if (props.token?.mint) {
-      return `https://jup.ag/swap/USDC-${props.token?.mint.toString()}?inAmount=1`;
+  const { wallet } = useWallet();
+  const connection = useWalletStore((s) => s.connection);
+  const endpoint = useMemo(() => connection.endpoint, [connection]);
+
+  const mint = useMemo(() => props.token?.mint.toString(), [props.token?.mint]);
+  const initJupiter = useCallback(() => {
```
nramadas, line 72: instead of this, i would import the scripts dynamically. You could do something like this:

```ts
interface Jupiter {
  init(args: any): any;
}

function importJupiter() {
  const script = new Promise((res, rej) => {
    const existing = document.getElementById(
      'jupiter-load-script',
    ) as HTMLScriptElement | null;

    if (existing) {
      existing.onload = res;
      existing.onerror = rej;
    } else {
      const el = document.createElement('script');
      el.onload = res;
      el.onerror = rej;
      el.id = 'jupiter-load-script';
      el.type = 'text/javascript';
      el.src = 'https://terminal.jup.ag/m ain.js';
      document.head.append(el);
    }
  });

  const css = new Promise((res, rej) => {
    const existing = document.getElementById(
      'jupiter-load-styles',
    ) as HTMLLinkElement | null;

    if (existing) {
      existing.onload = res;
      existing.onerror = rej;
    } else {
      const el = document.createElement('link');
      el.onload = res;
      el.onerror = rej;
      el.id = 'jupiter-load-styles';
      el.rel = 'stylesheet';
      el.href = 'https://terminal.jup.ag/main.css';
      document.head.append(el);
    }
  });

  return Promise.all([script, css]).then(() => {
    return (window as any).Jupiter as Jupiter;
  });
}
```

then, in the click handler, you could do the following:

```ts
onClick={async () => {
  const jupiter = await importJupiter();
  jupiter.init({
    /* ... */
  }) 
}}
```

hub/components/RealmHeader/index.tsx
```diff
@@ -11,13 +11,17 @@ import OverflowMenuHorizontalIcon from '@carbon/icons-react/lib/OverflowMenuHori
 import ProgressBarRound from '@carbon/icons-react/lib/ProgressBarRound';
 import UserFollow from '@carbon/icons-react/lib/UserFollow';
 import WalletIcon from '@carbon/icons-react/lib/Wallet';
+
```
nramadas, line 14: could we remove this blank line?

nramadas (changes requested):
Left a couple of comments, thanks!

hub/components/RealmHeader/index.tsx
dicksonpys: certainly!

pages/_document.tsx
dicksonpys: Reason I didnt load it dynamically in the original commit is because there's plan to support on-demand loading of Jupiter

Since the PR is targeting main.js it will always get the latest version of the script.

but I've updated it to use your code snippet, which works great.

because we support different displayModes, I needed some time to think about some details regarding integrated/widget mode.

hub/components/RealmHeader/index.tsx
dicksonpys: works great!

nramadas (approved):

PR #1296, opened 2022-12-13
fix default values after upgrade

abrzezinski94:


stores/useWalletStore.tsx
```diff
@@ -368,28 +370,41 @@ const useWalletStore = create<WalletStore>((set, get) => ({
...
         getRealmConfigAccountOrDefault(connection, programId, realmId),
       ])
-
-      //during the upgrade from v2 to v3 some values are undefined we need to ensure the defaults that match the program: 10 for depositExemptProposalCount and 0 for votingCoolOffTime
+      //during the upgrade from v2 to v3 some values are undefined
+      //we need to ensure the defaults that
+      //match the program:
+      //10 for depositExemptProposalCount and 0 for votingCoolOffTime
+      const governancesToSetDefaultValues = governances
+        .filter((x) => x.account.config.councilVoteThreshold.value === 0)
```
SebastianBor, line 378: the condition should also check for `councilVoteThreshold.type === VoteThresholdType.YesVotePercentage`

PR #1313, opened 2022-12-22
Feat/sns domains as destination

0xcen:
Adds support for .sol domains as destination addresses in Mint and Transfer SPL token transactions. 

todo: As it stands, tokenized domains don't resolve properly (as seen on the failing test) yet. I suspect a dependency conflict  is the problem. 

I've tested @bonfida/spl-name-service@0.1.58 & 0.1.59 and verified they do resolve tokenized domains properly, although I've had issues when bumping the version here. 

Screenshot 2022-12-21 at 15 22 47


pages/dao/[symbol]/proposal/components/instructions/SplTokenTransfer.tsx
```diff
@@ -116,9 +117,20 @@ const SplTokenTransfer = ({
...
       debounce.debounceFcn(async () => {
-        const pubKey = tryParseKey(form.destinationAccount)
+        let pubKey: PublicKey | null = null
+        if (destinationAddress.trim().toLowerCase().endsWith('.sol')) {
+          pubKey = await tryParseDomain(destinationAddress)
+        } else {
+          pubKey = tryParseKey(destinationAddress)
+        }
+
+        handleSetForm({
```
Comment (asktree) [inline pages/dao/[symbol]/proposal/components/instructions/SplTokenTransfer.tsx:129]: We should find a way to code this declaratively instead of imperatively.

PR #1315, opened 2022-12-27
Jupiter terminal: adjust terminal position for mobile

dicksonpys:
This PR adjust the Terminal's position so that it isn't blocked by the nav header. And more accessible on landscape mode.

From:
image
image

To:
image
image

### DIFF
--- a/hub/components/RealmHeader/index.tsx
+++ b/hub/components/RealmHeader/index.tsx
@@ -84,25 +84,7 @@ async function importJupiter() {
     }
   });
 
-  const css = new Promise((res, rej) => {
-    const existing = document.getElementById(
-      'jupiter-load-styles',
-    ) as HTMLLinkElement | null;
-
-    if (existing) {
-      res({});
-    } else {
-      const el = document.createElement('link');
-      el.onload = res;
-      el.onerror = rej;
-      el.id = 'jupiter-load-styles';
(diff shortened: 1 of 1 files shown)


hub/components/RealmHeader/index.tsx
```diff
@@ -84,25 +84,7 @@ async function importJupiter() {
...
-      el.onerror = rej;
-      el.id = 'jupiter-load-styles';
-      el.rel = 'stylesheet';
-      el.href = 'https://terminal.jup.ag/main.css';
-      document.head.append(el);
-    }
-  });
-
-  return Promise.all([script, css]).then(() => {
+  return Promise.all([script]).then(() => {
```
GregoryNEUT, line 87: nit: no need for `Promise.all` anymore now that there is only one promise

GregoryNEUT (commented):
Hello, just a nit :) hope it helps

hub/components/RealmHeader/index.tsx
dicksonpys: this is done!

dicksonpys: also, I've noticed that user's aren't actually auto connected on refresh some times, I've also pushed a conditional that only does passThrough when it is connected.
```tsx
passThroughWallet: connected ? wallet : undefined,
```

PR #1360, opened 2023-01-30
Wizards deposit council tokens directly to DAO

asktree:
Previously the wizards still used the old way of adding members: mint tokens to them and then they can deposit. This PR makes the wizards instead deposit council membership tokens directly into DAO, a v3 feature. 

Example DAO using this PR: http://localhost:3000/dao/3eT6tChqwZS67cXj1giWXXYDaJ7Pxm8Qr5oEWYYVgDum?cluster=devnet

### DIFF
--- a/tools/governance/prepareRealmCreation.ts
+++ b/tools/governance/prepareRealmCreation.ts
@@ -188,24 +188,16 @@ export async function prepareRealmCreation({
   // Create council mint
   let councilMintPk
 
-  // TODO explain the circumstances under which this set of conditions would be true
   if (
     zeroCommunityTokenSupply &&
     zeroCouncilTokenSupply &&
     nftCollectionCount === 0 &&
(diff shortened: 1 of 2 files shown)


tools/governance/prepareRealmCreation.ts
```diff
@@ -187,24 +188,16 @@ export async function prepareRealmCreation({
...
-      connection,
-      mintsSetupInstructions,
-      mintsSetupSigners,
-      walletPk,
-      null,
-      0,
-      walletPk
-    )
-  } else if (!existingCouncilMintPk && createCouncil) {
+    throw new Error('no tokens exist that could govern this DAO')
```
asktree, line 197: if this function is told to do things we don't want it to do, it should error, instead of silently attempting to find the best solution.

SebastianBor (approved):
Perfect

PR #1363, opened 2023-02-01
AddMemberForm updated to v3 membership flow

asktree:
In v3 we want to deposit council membership tokens directly into the DAO, so this updates the AddMemberForm instructions when using v3 or higher.

here is an example proposal created using this branch: https://app.realms.today/dao/2FbZcm8dyy1Pb9cJuvXCux5AsbF71XAEbzPEnHg4SAk1/proposal/FrLJzCnCgS8Bgv5ZJBePLmn1CA9CcSEXMBmwW8Ujmhh4?cluster=devnet

Note the above proposal cannot be signed due to an SDK bug https://github.com/solana-labs/oyster/pull/576/files -- the recipient is a signer. But if you simulate it you can see it works.

### DIFF
--- a/actions/executeInstructions.ts
+++ b/actions/executeInstructions.ts
@@ -5,7 +5,11 @@ import {
   RpcContext,
   withExecuteTransaction,
 } from '@solana/spl-governance'
-import { Transaction, TransactionInstruction } from '@solana/web3.js'
+import {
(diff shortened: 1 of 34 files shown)


asktree:
For reasons unclear to me this includes commits from merged PR https://github.com/solana-labs/governance-ui/pull/1360

PR #1365, opened 2023-02-01
AddMemberForm to v3 + UI changes for membership assets

asktree:
This PR also merges branch from #1363, I combined them while testing and refactoring some stuff.

---
image
This PR implements the following for membership mint assets: 
- Change label to *Council/Community Membership*
- Rename *supply* to *member(s)*
- Change icon to *Users*

Currently it also shows these changes for any council mint asset if you're using v2, though I have some concerns about this confusing users about the difference between membership and liquid. 

I also refactor `AddMemberForm` to work if community is `Membership`.

Example of successful membership adding: http://localhost:3000/dao/2FbZcm8dyy1Pb9cJuvXCux5AsbF71XAEbzPEnHg4SAk1/proposal/D17BUtM7JBfWih5qDnogK5k4rEPDRpbdv1odB3xXW4am?cluster=devnet


asktree:
Failing to build due to a bizarre TS issue, here is a post I wrote in TS discord 
image

asktree:
TS versioning issue. whatever

SebastianBor:
When we select `Mint Tokens` from the W&A view 

image

Then the proposal form should pre-populate the relevant mint 

image

And I think for the council token we can also set amount to 1 by default 

image

SebastianBor:
When the proposal to mint membership tokens is approved then we should show that only the target wallet can execute it.
- Disable the `Execute` button for other users 
- Show info explaing that only the given wallet can execute it 

image

The general use case for this is: If there is any signer in the instruction which is not the governance or DAO wallet address then we should apply the logic

SebastianBor:
This warning is incorrect. The program is known. It's the spl-gov which the DAO uses 

image

PR #1379, opened 2023-02-05
[stacked] Revoke membership flow from Members page

asktree:
Adds a button to remove members in the Members view when using V3, which opens a modal to make a proposal.
image

Example of successful proposal using this flow:
http://localhost:3000/dao/2FbZcm8dyy1Pb9cJuvXCux5AsbF71XAEbzPEnHg4SAk1/proposal/6SDMuvSoJY88q48mmcwkz6JcWKx2psgNVt9CCNZ2EXp1?cluster=devnet


asktree:
Closing because @SebastianBor and I discussed and decided to minimize codebase by not including this flow and instead only use generic proposal form, given that this flow will see so little usage. Work is complete and can be reopened later.

PR #1389, opened 2023-02-08
[quickie] fix side effect bug in governance power card

asktree:
After using a proposal to revoke my membership in the DAO, I was getting an bug in a useEffect where a `!`'d `realm` object was actually undefined. I didn't really diagnose this issue, I just made everything type safe and made the deps exhaustive, and the issue seems to be resolved.


asktree:
No idea why a test involving pubkeys is failing

PR #1390, opened 2023-02-08
Revoke Membership instruction for generic proposal flow

asktree:
image

successful example: http://localhost:3000/dao/2FbZcm8dyy1Pb9cJuvXCux5AsbF71XAEbzPEnHg4SAk1/proposal/2Dhq7d5CncnjEjswgWbT5HuF2kaQFbSqhLx7gb5hrsrF?cluster=devnet


SebastianBor:
Looks good, the only issue I noticed is after creating a proposal it stays on the proposal screen and looks like nothing happened. Once the proposal is created it should navigate back to the proposals view

SebastianBor:
The back button points to the proposal view itself and doesn't work 

image

asktree:
@SebastianBor I am pretty confident that these issues are not downstream of this PR? But I will ask @abrzezinski94

PR #1394, opened 2023-02-09
Revoke membership flows

asktree:
Adds 2 flows that prefill the revoke membership form:
image
image

Clicking either button takes you to a new proposal form with pre-filled information.


SebastianBor:
When current user has no governance power to create the revoke proposal then the button should be disabled 

image

SebastianBor:
When revoking a specific member let's also pre set the title. Ex: `'Revoke EaWH...Fkta6 Council memebership'`

image

asktree:
I made the changes requested

PR #1398, opened 2023-02-10
Treasury trade UI

tomjohn1028:
* Remove serum remote naming & package for Poseidon
* Integrate V2 instruction set for Poseidon
* More user friendly interface
v2_governance_modal


tomjohn1028:
@SebastianBor This is in generally in a state that I'd say is code complete. There are two items before I mark as read though.

1. Is it acceptable that this adds the `react-datetime-picker` package? I did not want to write something custom. 
2. With the added dependency, it's causing tests to fail because of nested node_modules in `@types/react-calendar` and `@types/react-dom`. I see that we ignore scripts, so I cannot use the post install script to clean up these files. Why are scripts being ignored?

components/instructions/tools.tsx
```diff
@@ -320,6 +322,7 @@ export async function getInstructionDescriptor(
   } else {
     descriptors = INSTRUCTION_DESCRIPTORS[instruction.programId.toBase58()]
   }
+  console.log('instruction descriptor', instruction.data[0])
```
abrzezinski94, line 325: nit: remove console log

abrzezinski94:
It would be good to remove react calendar and just use native https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local if it would not create a lot of work

PR #1408, opened 2023-02-16
Add a new form for editing the Realm Config

nramadas:


components/treasuryV2/Details/RealmAuthorityDetails/Config.tsx
```diff
@@ -248,12 +252,6 @@ export default function Config(props: Props) {
         )}
       </div>
       <NFTVotePluginSettingsDisplay className="mt-24" />
-      {editRealmOpen && (
-        <RealmConfigModal
```
asktree, line 252: I think it would be best to delete this now unused modal from the codebase entirely. This will also tell you if it's used anywhere that you missed.

pages/dao/[symbol]/params/components/ParamsView.tsx
```diff
@@ -137,7 +141,9 @@ const ParamsView = ({ activeGovernance, openGovernanceProposalModal }) => {
               tooltipMessage={
                 'Please connect wallet with enough voting power to create governance config proposals'
               }
-              onClick={openGovernanceProposalModal}
```
Comment (asktree) [inline pages/dao/[symbol]/params/components/ParamsView.tsx:140]: This button opens the `GovernanceConfigModal` which edits governance config, not realm config. But you are replacing it with a link to the realm config form you made. So that seems like a bug.

asktree (changes requested):

PR #1458, opened 2023-03-09
set more secure defaults when making governances

asktree:


components/TreasuryAccount/NewTreasuryAccountForm.tsx
```diff
@@ -53,7 +53,7 @@ const defaultFormValues = {
   minInstructionHoldUpTime: 0,
   maxVotingTime: 3,
   voteThreshold: 60,
-  voteTipping: VoteTipping.Strict,
```
SebastianBor, line 56: - The UI should ask for 'Council Vote Tipping' and set it accordingly (please update label in the form as well). The default value in the form should be `Strict` 
- The Community Vote Tipping should always be set to `Disabled`

asktree:
these views have all been replaced instead

PR #1478, opened 2023-03-17
HeliumVotePlugin

bryzettler:
# HeliumVotePlugin
adds support for the [Helium vsr instance](https://explorer.solana.com/address/hvsrNC3NKbcryqDs2DocYHZ9yPKEVzdSjQG6RVtK1s8?cluster=devnet) and Ux/Ui for how the Helium DAO, Helium IOT DAO, Helium MOBILE DAO will be utilized.

## Some callouts.
- Piggybacked our `initializeRegistrar` and `configVotingMint` with existing forms. We have scripts that set these up so we could possibly drop those from the code altogether if its a anti pattern
- Should replace the `metaplex.nfts()` logic in `getPositions` and `getSubDaos` once moved off Holaplex
- Was my intention to have `LockTokensModal.tsx` be reusable for both VoteStakeRegistry and HeliumVotePlugin, Changes were happening in parallel to upstream remote so focused on our functionality first. Can revisit later


bryzettler:
Ignore the vercel deploy request. Not sure why thats here

.nvmrc
```diff
@@ -1 +1 @@
-lts/fermium
+lts/gallium
```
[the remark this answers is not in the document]
bryzettler: While deploying to cloudflare pages it would pull the node version from here. fermium is a 14.x and some of our packages depend on >=16.x. So bumped it to gallium which is that.

utils/truthy.ts
```diff
@@ -0,0 +1,2 @@
+export type Truthy<T> = T extends false | '' | 0 | null | undefined ? never : T // from lodash
```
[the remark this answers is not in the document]
bryzettler: Yeah fair, we use this as a util somewhere else in our code so pulled it in. I nixed the implementation and usage of it.

utils/validations.tsx
```diff
@@ -981,7 +983,9 @@ export const getRealmCfgSchema = ({
               if (val) {
                 try {
                   getValidatedPublickKey(val)
-                  if ([...nftPluginsPks].includes(val)) {
+                  if (
```
asktree, line 986: should this array not just be `supportedPlugins`?
@abrzezinski94?

utils/validations.tsx
```diff
@@ -1118,7 +1122,9 @@ export const getRealmCfgSchema = ({
...
```
asktree, line 1125: same comment as other array

utils/tokens.tsx
```diff
@@ -439,27 +444,51 @@ export const getNfts = async (
   return await getNftsFromHolaplex(ownerPk, connection)
 }
 
+// TODO: Remove when upgrade spl-token
```
asktree, line 447: What's the upgrade referred to here? Sorry if I'm out of the loop

utils/tokens.tsx
```diff
@@ -485,3 +514,7 @@ export function getScaledFactor(amount: number) {
     new BigNumber(amount.toString()).shiftedBy(SCALED_FACTOR_SHIFT).toString()
   )
 }
+
+export function getUnscaledFactor(amount: number) {
```
asktree, line 518: I find this util confusing:
- it seems to be the inverse of `getScaledFactor` but accepts a `number` instead of `BN`
- to be pedantic, `getInverseScaledFactor` seems more accurate. 
- I also don't like utils that say `getFoo` when they don't fetch anything and instead compute, I'd rather say `computeFoo`, but I recognize there's a precedent set by the sister util here.

utils/dateTools.ts
```diff
@@ -52,18 +55,16 @@ export const daysToMonths = (days: number) => {
   return days / DAYS_PER_MONTH
 }
 
-export const getMinDurationFmt = (deposit: DepositWithMintAccount) => {
-  return getFormattedStringFromDays(getMinDurationInDays(deposit))
+export const getMinDurationFmt = (startTs: BN, endTs: BN) => {
```
asktree, line 58: great change

utils/uiTypes/VotePlugin.ts
```diff
@@ -281,12 +353,138 @@ export class VotingClient {
...
+        clientProgramId
+      )
+
+      const [maxVoterWeightPk] = maxVoterWeightRecordKey(
+        realm.pubkey,
+        realm.account.communityMint,
+        clientProgramId
+      )
+
+      for (let i = 0; i < unusedPositions.length; i++) {
```
asktree, line 436: style nit: consider `remainingAccounts = unusedPositions.map(blahblah)`

stores/useVotePluginsClientStore.tsx
```diff
@@ -74,11 +92,13 @@ interface UseVotePluginsClientStore extends State {
 
 const defaultState = {
   vsrClient: undefined,
+  heliumVsrClient: undefined,
   nftClient: undefined,
   gatewayClient: undefined,
   switchboardClient: undefined,
   pythClient: undefined,
   voteStakeRegistryRegistrar: null,
+  heliumVsrRegistrar: null,
```
asktree, line 101: (dont change) I'm amused by the mixed usage of undefined and null in the existing state, ha

pages/dao/[symbol]/proposal/components/instructions/Vsr/VotingMintConfig.tsx
```diff
@@ -28,18 +27,25 @@ import {
...
-  grantAuthority: AssetAccount | undefined
   mintDigitShift: number
-  mintUnlockedFactor: number
-  mintLockupFactor: number
-  mintLockupSaturation: number
+  maxLockupFactor: number
+  lockupSaturation: number
+
+  // Helium DynamicFields based on programId
+  baselineVoteWeightFactor?: number
```
Comment (asktree) [inline pages/dao/[symbol]/proposal/components/instructions/Vsr/VotingMintConfig.tsx:44]: i would like it if these all lived grouped together in one object. 
so `helium.baselineVoteWeightFactor` instead of `baselineVoteWeightFactor`. This would also let you be more typesafe.

pages/dao/[symbol]/proposal/components/instructions/Vsr/VotingMintConfig.tsx
```diff
@@ -52,76 +58,147 @@ const VotingMintConfig = ({
...
+    if (form?.programId) {
+      if (vsrPluginsPks.includes(form.programId)) {
+        setShowHeliumFields(false)
+      }
+
+      if (heliumVsrPluginsPks.includes(form.programId)) {
+        setShowHeliumFields(true)
+      }
+    }
+  }, [form?.programId, setShowHeliumFields])
```
Comment (asktree) [inline pages/dao/[symbol]/proposal/components/instructions/Vsr/VotingMintConfig.tsx:77]:
```suggestion
  const showHeliumFields = useMemo(() => heliumVsrPluginsPks.includes(form.programId) 
  }, [form?.programId, vsrPluginPks heliumVsrPluginsPks])
```
Would this work? Seems like this should be a memoized computation, and not be in state.

tsconfig.json
```diff
@@ -15,6 +15,7 @@
     "resolveJsonModule": true,
     "jsx": "preserve",
     "paths": {
+      "@solana/*": ["node_modules/@solana/*"],
```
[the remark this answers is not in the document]
bryzettler: I think my imports from '@solana` were erroring out prior to adding this. Ill remove and see whats up.

utils/tokens.tsx
bryzettler: Ah this is left over, I was going to upgrade spl-token but ended up not having to.

pages/dao/[symbol]/proposal/components/instructions/Vsr/VotingMintConfig.tsx
```diff
@@ -192,17 +264,25 @@ const VotingMintConfig = ({
           }
         }
       ),
+    ...(!showHeliumFields
```
Comment (asktree) [inline pages/dao/[symbol]/proposal/components/instructions/Vsr/VotingMintConfig.tsx:267]: nit: i think its weird how the schema lives inside this component. thats not your fault but would it be possible to make this check simply look at the form contents, rather than depending on external state? and then maybe just extract this schema? or at least memoize it.

pages/dao/[symbol]/proposal/components/instructions/Vsr/CreateRegistrar.tsx
```diff
@@ -42,73 +63,102 @@ const CreateVsrRegistrar = ({
...
   }
+
   useEffect(() => {
     handleSetInstructions(
       { governedAccount: form?.governedAccount?.governance, getInstruction },
       index
     )
     // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO please fix, it can cause difficult bugs. You might wanna check out https://bobbyhadz.com/blog/react-hooks-exhaustive-deps for info. -@asktree
   }, [form])
-  const schema = yup.object().shape({
```
Comment (asktree) [inline pages/dao/[symbol]/proposal/components/instructions/Vsr/CreateRegistrar.tsx:90]: good

pages/dao/[symbol]/proposal/components/instructions/Vsr/CreateRegistrar.tsx
```diff
@@ -42,73 +63,102 @@ const CreateVsrRegistrar = ({
...
-        }
-        return true
-      })
-      .required('VSR Program ID is required'),
-  })
+
   const inputs = [
     {
       label: 'Wallet',
-      initialValue: null,
```
Comment (asktree) [inline pages/dao/[symbol]/proposal/components/instructions/Vsr/CreateRegistrar.tsx:111]: good

pages/dao/[symbol]/proposal/components/instructions/Validators/StakeValidator.tsx
```diff
@@ -203,24 +203,14 @@ const StakeValidator = ({
...
-    handleSetInstructions(
-      {
-        governedAccount: governedAccount,
-        getInstruction,
-      },
-      index
-    )
-    // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO please fix, it can cause difficult bugs. You might wanna check out https://bobbyhadz.com/blog/react-hooks-exhaustive-deps for info. -@asktree
-  }, [form])
-
```
Comment (asktree) [inline pages/dao/[symbol]/proposal/components/instructions/Validators/StakeValidator.tsx:216]: lol

pages/dao/[symbol]/account/[tokenOwnerRecordPk].tsx
```diff
@@ -1,32 +1,38 @@
...
+import { vsrPluginsPks, heliumVsrPluginsPks } from '@hooks/useVotingPlugins'
 import { useRouter } from 'next/router'
-import Account from 'VoteStakeRegistry/components/Account/Account'
+import { default as VoteStakeRegistryAccount } from 'VoteStakeRegistry/components/Account/Account'
 import LockTokensAccount from 'VoteStakeRegistry/components/Account/LockTokensAccount'
+import { LockTokensAccount as HeliumLockTokensAccount } from 'HeliumVotePlugin/components/LockTokensAccount'
 
-const account = () => {
-  // eslint-disable-next-line react-hooks/rules-of-hooks -- TODO this is potentially quite serious! please fix next time the file is edited, -@asktree
+const Account: React.FC<any> = () => {
```
Comment (asktree) [inline pages/dao/[symbol]/account/[tokenOwnerRecordPk].tsx:9]: shouldn't it be just `React.FC`? no props

pages/dao/[symbol]/account/[tokenOwnerRecordPk].tsx
```diff
@@ -1,32 +1,38 @@
...
-    vsrPluginsPks.includes(
-      config?.account.communityTokenConfig.voterWeightAddin?.toBase58()
-    )
+  const communityVsrPk = config?.account.communityTokenConfig.voterWeightAddin
+  const isDefaultVsrMode =
+    communityVsrPk && vsrPluginsPks.includes(communityVsrPk.toBase58())
+  const isHeliumVsrMode =
+    communityVsrPk && heliumVsrPluginsPks.includes(communityVsrPk.toBase58())
 
-  const getAccountView = () => {
```
Comment (asktree) [inline pages/dao/[symbol]/account/[tokenOwnerRecordPk].tsx:19]: hmmmmmmmm

hub/providers/Proposal/fetchPlugins.ts
```diff
@@ -1,4 +1,4 @@
-import { AnchorProvider, Wallet } from '@project-serum/anchor';
```
asktree, line 1: I'm surprised this compiled?

hooks/useVotingPlugins.ts
```diff
@@ -60,70 +70,108 @@ export function useVotingPlugins() {
...
-    setIsLoading,
-    setVotingPower,
-    setOracleKeys,
-    setInstructions,
-  } = useSwitchboardPluginStore()
 
+  const nftStore = useNftPluginStore()
+  const heliumStore = useHeliumVsrStore()
+  const gatewayStore = useGatewayPluginStore()
+  const switchboardStore = useSwitchboardPluginStore()
```
asktree, line 77: good

asktree: i hate this file

utils/uiTypes/VotePlugin.ts
```diff
@@ -324,94 +518,38 @@ export class VotingClient {
...
-        'castVote'
-      )
-      return props
-    }
-    if (this.client instanceof SwitchboardQueueVoterClient) {
-      const props = await this.withUpdateVoterWeightRecord(
-        instructions,
-        tokeOwnerRecord,
-        'castVote'
+      const remainingNftsChunk = chunks(
```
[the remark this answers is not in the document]
bryzettler: Yeah my apologies, Since our implementation of vsr uses NFTs I used the current implementation of NFTVoterClient and mirrored a-lot of the logic. When it came to this function it seemed redundant to just call the same method on the program `castNftVoteIx2 ` when there was already a loop above it going over the chunks and calling that instruction. So I just combined them with for `(const chunk of [firstFiveNfts, ...remainingNftsChunk])`. The testing I would say would be done through our vsr instance as its using NFTs. I can go ahead and create a mock realms instance with some fake nfts to test if you want.

hooks/useVotingPlugins.ts
asktree: would it be possible to make all Helium stuff live in its own hook, which just returns `handleHeliumBlahblah` and then have this abominable hook simply use that?

utils/tokens.tsx
bryzettler: ah yeah, fair callout

hooks/useRealm.tsx
```diff
@@ -258,7 +265,8 @@ export default function useRealm() {
     ownDelegateCouncilTokenRecords,
     config,
     currentPluginPk,
-    isLockTokensMode,
+    isDefaultVsrMode,
```
asktree, line 268: how annoying would it be to move `isDefaultVsrMode` and `isHeliumVsrMode` to their own hooks? I don't know who invented the mega-hook but I want to abolish it.

hooks/useNft.ts
```diff
@@ -0,0 +1,41 @@
...
+import {
+  keypairIdentity,
+  Metaplex,
+  Nft,
+  NftWithToken,
+  Sft,
+  SftWithToken,
+} from '@metaplex-foundation/js'
+
+export const useNft = (mint: PublicKey | undefined) => {
```
asktree, line 13: nit: this seems like it wants to live a happy life as a `react-query` query. you can check the `queries` folder for reference. I will not enforce this change but I like `react-query` and suggest this.

utils/uiTypes/VotePlugin.ts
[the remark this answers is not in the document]
asktree: I agree seems redundant, the original code is just hard to follow and I dont feel confident in my ability to audit. Testing would be good

tsconfig.json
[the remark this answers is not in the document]
asktree: Thats very strange. Maybe a symlink thing? Let's remove this and if it doesn't work I can help try to debug your environment.

components/ProposalMyVoteBadge.tsx
```diff
@@ -52,29 +52,49 @@ interface Props {
...
+    communityDelegateVoteRecords,
+    councilDelegateVoteRecords,
+  ] = useWalletStore((s) => [
+    s.ownVoteRecordsByProposal,
+    s.communityDelegateVoteRecordsByProposal,
+    s.councilDelegateVoteRecordsByProposal,
+  ])
 
-  const ownVoteRecord = getOwnVoteRecord(
+  useEffect(() => {
```
asktree, line 69: ownVoteRecord should be memoized using `useMemo`, not state and effect.

components/DropdownBtn.tsx
```diff
@@ -1,72 +1,101 @@
...
-  if (!defaultFunction) {
-    throw 'DropdownBtn must have at least one default option'
-  }
-  const filtredOptions = options.filter((x) => !x.isDefault)
+  const basicClasses = `inline-flex rounded-full w-full justify-center bg-primary-light text-bkg-2 px-4 py-3 text-sm font-medium`
+  const hoverClasses = `hover:bg-opacity-30 focus:outline-none focus-visible:ring-2 focus-visible:ring-white focus-visible:ring-opacity-75 hover:bg-fgd-1`
+  const disabledClasses = disabled
+    ? `!bg-fgd-4 !cursor-not-allowed !text-fgd-3`
+    : ''
+  const btnClasses = `${basicClasses} ${hoverClasses} ${disabledClasses}`
```
asktree, line 35: nit: hmm. my ide has autocomplete and other nice features for tailwind classes. Instead of breaking these up in this manner, I would suggest simply having the classnames inline, but break them up with line breaks and comments.

asktree: on the other hand i don't care so you can ignore this. this component is surely not long for this world anyway.

components/Button.tsx
```diff
@@ -21,19 +22,21 @@ const Button: FunctionComponent<ButtonProps> = ({
...
   tooltipMessage = '',
   style,
+  type = 'button',
   ...props
 }) => {
   return (
     <button
       className={`${className} default-transition font-bold px-4 rounded-full ${
-        small ? 'py-1' : 'py-2.5'
+        small ? 'py-2' : 'py-2.5'
```
asktree, line 31: messing with style of core components makes me nervous, can you comment? How can we be confident it doesn't mess stuff up elsewhere?

components/App.tsx
```diff
@@ -77,7 +76,7 @@ export function App(props: Props) {
   const { realm, ownTokenRecord, realmInfo, symbol, config } = useRealm()
   const wallet = useWalletStore((s) => s.current)
   const connection = useWalletStore((s) => s.connection)
-  const client = useVotePluginsClientStore((s) => s.state.vsrClient)
+  const [vsrClient] = useVotePluginsClientStore((s) => [s.state.vsrClient])
```
asktree, line 79:
```suggestion
  const vsrClient = useVotePluginsClientStore((s) => s.state.vsrClient)
```

components/VotePanel/YouVoted.tsx
```diff
@@ -179,7 +179,7 @@ export const YouVoted = ({ quorum }: { quorum: 'electoral' | 'veto' }) => {
               onClick={() => submitRelinquishVote()}
               disabled={!isWithdrawEnabled || isLoading}
             >
-              Withdraw
+              Relinquish Vote
```
asktree, line 182: Wouldn't be surprised if this causes awkward linebreaks at certain window widths, but if you dont think so I'll take your word for it.

pages/dao/[symbol]/proposal/components/instructions/Vsr/VotingMintConfig.tsx
Comment (bryzettler) [reply pages/dao/[symbol]/proposal/components/instructions/Vsr/VotingMintConfig.tsx:44]: actually, thanks for calling this out. In our vsr we used to require a `minimumLockupUpTs` but now we dont and baselineVoteWeightFactor is shared across both vsr instances. The only actual logic happening here is hiding/showing grantAuth if the programId isnt helium. Don't even need these fields anymore

utils/uiTypes/VotePlugin.ts
```diff
@@ -178,11 +198,63 @@ export class VotingClient {
       return { voterWeightPk, maxVoterWeightRecord: undefined }
     }
 
+    if (this.client instanceof HeliumVsrClient) {
```
asktree, line 201: nit: not your fault but SURELY the way this stuff should be done, is that you would have a common interface for a `VotingClient`, and then each type of voting client implements that interface. NOT one giant class with a ton of if statements. 
Would it be too much work for you to adopt this approach for Helium? Extract whatever the API of this VotingClient is, make an interface for it, and then have `MegalithicOmniVotingClient` which is the current one, and `HeliumVotingClient`. And then whenever in the stores the voting client is set, it decides which to instantiate.

components/ProposalVotingPower/LockedCommunityNFTRecordVotingPower.tsx
```diff
@@ -0,0 +1,215 @@
...
+    }
+  }, [positions, currentClient, propagatePositions])
+
+  useEffect(() => {
+    if (mint && votingPower) {
+      setAmount(getMintDecimalAmount(mint, votingPower))
+    }
+  }, [mint, votingPower])
+
+  useEffect(() => {
```
asktree, line 69: there is a query for exactly this in the queries folder, I suggest using it. I've seen this copy pasted around the codebase and it has race conditions (which you may have fixed, I didnt bother to look super close).

asktree: useAddressQuery_TokenOwnerRecord, useAddressQuery_CommunityTokenOwner, useAddressQuery_CouncilTokenOwner

hub/providers/Proposal/fetchPlugins.ts
bryzettler: It did for me 🤷‍♂️

actions/dryRunInstruction.ts
```diff
@@ -15,6 +16,11 @@ export async function dryRunInstruction(
   additionalInstructions?: InstructionData[]
 ) {
   const transaction = new Transaction({ feePayer: wallet.publicKey })
+
+  transaction.add(
```
asktree, line 20: Is this desireable? Might it cause transactions to succeed in dry run that would fail in the real world? The answer may be no, I have no idea.

hooks/useNft.ts
bryzettler: I dont think I even use this file anywhere, going to kill it. It was goign to die with helius anyways.

components/VotePanel/YouVoted.tsx
bryzettler: don't think it does. Ran through all devices in the chrome inspector and looked like it had ample room

actions/dryRunInstruction.ts
bryzettler: compute unit is already bumped to 1_000_000 in both executeInstructions.ts and executeTransaction.ts. There might be a better way of doing this but if we didn't have it here thinks would fail which would otherwise pass in both of those.

hooks/useRealm.tsx
bryzettler: What would it be, useVsrMode? that just returns a enum.

actions/castVote.ts
```diff
@@ -140,36 +141,87 @@ export async function castVote(
...
+
+  if (!isNftVoter && !isHeliumVoter) {
+    const transaction = new Transaction()
+    transaction.add(...instructions)
+
+    await sendTransaction({ transaction, wallet, connection, signers })
+  }
+
+  // we need to chunk instructions
+  if (isHeliumVoter) {
```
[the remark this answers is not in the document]
bryzettler: Honestly I saw the implementation for castVote for the nftplugin and just copypasta'd it here. I can revisit this if you want. I saw `message ? 4 : 2` for the chunking and Didn't want to dive into it since there were no comments around.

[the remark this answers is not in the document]
abrzezinski94: its a bit problematic to create generic chunk function some instructions have to be called in one chunk others not + chunking depends on size of single instruction.

hooks/useVotingPlugins.ts
bryzettler: Yeah this file is rough, that could be possible yeah. Is that something that would prevent this pr?

utils/uiTypes/VotePlugin.ts
bryzettler: Im not sure if I'll have the time to refactor our voting client to that, if I find it I can give it a go.

actions/castVote.ts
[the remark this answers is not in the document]
abrzezinski94: In case of nfts there is const instructionsCountThatMustHaveTheirOwnChunk = message ? 4 : 2
Because first chunk have to have certain instructions together in one chunk for example if you just cast vote then first chunk is just 2 castVote + withCastPluginVote but if you post msg with your vote we need to have 4 instructions together cast vote + update voter weight + msg + withCastPluginVote then rest i chunked by 2 as i remember.

so in some cases you have chunks where you have [4 inst, 2 inst, 2 inst] in other there is just [2inst, 2inst, 2inst]
 
But i should leave comment there and probably rename the variable because its only chunk of the first instructions that is measured there.

hooks/useVotingPlugins.ts
asktree: No its not a blocker, nothing regarding code quality really is much of a blocker.

hooks/useRealm.tsx
asktree: yeah sounds good (though don't use an actual enum, just a literal union `"helium" | "default" | undefined`)

actions/dryRunInstruction.ts
asktree: Got it. sounds good.

hooks/useRealm.tsx
```diff
@@ -27,11 +27,14 @@ import {
...
   pythPluginsPks,
   gatewayPluginsPks,
+  heliumVsrPluginsPks,
 } from './useVotingPlugins'
 import useGatewayPluginStore from '../GatewayPlugin/store/gatewayPluginStore'
 import useSwitchboardPluginStore from 'SwitchboardVotePlugin/store/switchboardStore'
+import useHeliumVsrStore from 'HeliumVotePlugin/hooks/useHeliumVsrStore'
 import { BN } from '@coral-xyz/anchor'
 import { PublicKey } from '@solana/web3.js'
+import { useVsrMode } from './useVsrMode'
```
asktree, line 37: to be clear, my intention was for `useVsrMode` to be used by components, rather than passing it through `useRealm` (since I don't like the way useRealm is just a mega-hook used for all sorts of disconnected data). But this is still step in right direction and I'm not going to bug you further about it. Thanks!

VoteStakeRegistry/sdk/accounts.tsx
```diff
@@ -76,6 +78,28 @@ export const getRegistrarPDA = async (
   }
 }
 
+export const getMaxVoterWeightPDA = async (
```
asktree, line 81: This PDA is helium specific right? if so I think it would be good to have a docstring stating this. If it is not then ignore this comment.

VoteStakeRegistry/components/Account/LockTokensAccount.tsx
```diff
@@ -44,7 +44,10 @@ interface DepositBox {
 }
 const unlockedTypes = ['none']
 
-const LockTokensAccount = ({ tokenOwnerRecordPk }) => {
+const LockTokensAccount: React.FC<{
+  tokenOwnerRecordPk: string | string[] | undefined
```
asktree, line 48: (not your fault, dont bother fixing) i appreciate that `tokenOwnerRecordPk` has many possible types, none of which are a PublicKey.

HeliumVotePlugin/components/LockCommunityTokensBtn.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const hasTokensInWallet =
+    realmTokenAccount && realmTokenAccount.account.amount.gt(new BN(0))
+
+  const tooltipContent = !connected
+    ? 'Connect your wallet to lock'
+    : !hasTokensInWallet
+    ? "You don't have any governance tokens in your wallet to lock."
+    : ''
+
+  const handleOnClick = () => {
```
asktree, line 28: nit: let's get rid of this fn

HeliumVotePlugin/components/LockTokensAccount.tsx
```diff
@@ -0,0 +1,410 @@
...
+        })
+      }
+    } catch (e) {
+      notify({
+        type: 'error',
+        message: 'Unable to fetch positions',
+      })
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- TODO please fix, it can cause difficult bugs. You might wanna check out https://bobbyhadz.com/blog/react-hooks-exhaustive-deps for info. -@asktree
+  }, [])
```
asktree, line 121: this looks dubious to me --
Firstly I notice this callback has no dependencies which seems like it could be a source of bugs. 
Secondly, if you use a callback like this with an effect, I think it's best to inline the callback inside of the effect. However, rather than using an effect, this seems like a perfect time to use the `react-async-hook` library.

HeliumVotePlugin/components/LockTokensModal.tsx
```diff
@@ -0,0 +1,362 @@
...
+      lockupKind: lockupKindOptions[0],
+      lockupPeriod: lockupPeriodOptions[0],
+      lockupPeriodInDays: lockupPeriodOptions[0].value,
+    },
+  })
+
+  const { lockupKind, amount, lockupPeriod, lockupPeriodInDays } = watch()
+
+  // validateAmountOnChange
+  useEffect(() => {
```
asktree, line 124: (non blocking) Rather than validating using effects, it would probably be best to use react-hook-form's API. I found this article https://jasonwatmore.com/post/2021/04/21/react-hook-form-7-form-validation-example 


The reason I say this is because validating (and modifying) stuff while the user is in the middle of editing it often results in annoying jank.

HeliumVotePlugin/components/PositionCard.tsx
```diff
@@ -0,0 +1,458 @@
...
+    undelegatePosition,
+  } = useUndelegatePosition()
+
+  const [
+    connection,
+    wallet,
+    { fetchRealm, fetchWalletTokenAccounts },
+  ] = useWalletStore((s) => [s.connection.current, s.current, s.actions])
+
+  useEffect(() => {
```
asktree, line 109: `transferablePositions` should use `useMemo` rather than an effect.

asktree (approved):
PR is approved pending some minor nits I left just now.
I appreciate Bry's patience and attention to detail. This PR leaves the codebase better than they found it. 

Things I didn't do in my review:
- Test anything at all.
- Closely business logic involving Helium-specific things.

Aside from Helium SDK, Bry introduced the `react-async-hook` package. Starting immediately I would like all network queries to make use of `react-query` unless they are hitting our web2 gateway, but I think `react-async-hook` is perfect for async PDA computation and I'd like us to keep it around for this purpose.

HeliumVotePlugin/components/LockTokensAccount.tsx
bryzettler: Good call out!

HeliumVotePlugin/components/LockTokensModal.tsx
bryzettler: Was actually doing some onchange side effects in this logic. Just moved to the actual on change of the field.

PR #1493, opened 2023-03-22
Uxd management

sjillen:
Add UXD interactions with UXD Protocol program to the list of instructions available in proposals.
Two Instructions are public and available for all DAOs:
- Mint and Redeem (usable with treasuries)
The rest of the instruction are dedicated to UXD Protocol management and are solely usable within UXD Protocol DAO.

This PR is one of the main step in regards of UXD Protocol DAO codebase reintegration into Realms.

This PR adds a new npm package dependency: `@uxd-protocol/uxd-client`


asktree:
I'd like to freeze all dao-related packages being added to the repo's package.json. Could you copy-paste the functions you need into a folder instead?

sjillen:
@asktree unfortunately, it's rather complicated to pull out just the functions we need because they depends on most of the 
code in the @uxd-protocol package

PR #1513, opened 2023-03-31
[UXD] Adding new flags to the edit depository IXs

uxd-vincent:


uxd-vincent:
Created by mistake in the wrong depository

PR #1522, opened 2023-04-04
Shared wallet API

asktree:
Currently we can't share components in both the hubs and governance apps due to different wallet providers. This creates a hook for using the Hub wallet api with the governance wallet. It also creates a HOC for injecting the hook's provider when needed. 

If we run into issues with the HOC I suggest inserting the providers at the root of each app.


asktree:
http://localhost:3000/dao/BTm6bfhAnGRgZ5gbZzCMBc7k1R3SMUCG6KLXhTnmW4sb/proposal/6so72jjHaozcNJF38A2NVt5svj88MEkoQxuQ5kyqTf8d?cluster=devnet here is a proposal I created using this branch

PR #1525, opened 2023-04-06
Updated the old Wezesha DAO with new details

dukemtambo:


dukemtambo (commented):
Updating the old Wezesha DAO with new details

dukemtambo (commented):
Hidding the old DAO

PR #1531, opened 2023-04-07
Use our rpc for WebBundlr

KishiTheMechanic:


KishiTheMechanic:
This is for fixing create/update token metadata modal

PR #1535, opened 2023-04-11
Helium: Display delegated subDao

bryzettler:
Adds a visual cue to the position card for which subDao its delegated to


asktree (approved):
lgtm in that I don't see anything that could break the rest of the app

PR #1539, opened 2023-04-13
set vsr client properly

bryzettler:


hooks/useVotingPlugins.ts
```diff
@@ -363,8 +363,12 @@ export function useVotingPlugins() {
   useEffect(() => {
     if (wallet?.publicKey?.toBase58()) {
       if (currentPluginPk) {
-        handleSetVsrClient(wallet, connection, currentPluginPk)
-        handleSetHeliumVsrClient(wallet, connection, currentPluginPk)
+        if (vsrPluginsPks.includes(currentPluginPk.toBase58())) {
```
asktree, line 366: im nervous about this because i dont know if `vsrPluginsPks` is exhaustive. Could we dynamically check if the account is a plugin instead?

@abrzezinski94 do you know if its exhaustive? will it be in future?

PR #1546, opened 2023-04-14
Upgrade `New Wallet` form

asktree:
Here is the upgraded New Wallet form.
I took the approach of recycling as many components as I could from @nramadas 's recent `Edit Rules` form. Along the way I made some existing components more generic and composed them into `Edit Rules` and `New Wallet` versions. 

Unlike `Edit Rules`, this lives in the normal 1.0 app hierarchy, which wasn't too much trouble really. 

In order to generate defaults, it looks at all existing governances and determines the strictest setting for each field. 

It looks like this 
image


hub/components/EditRealmConfig/Summary/index.tsx
```diff
@@ -45,11 +45,11 @@ export function Summary(props: Props) {
       />
       <ProposalVoteType
         className="mt-8"
-        currentCommunityRules={props.governance.communityTokenRules}
-        currentCouncilRules={props.governance.councilTokenRules}
-        currentBaseVoteDays={props.governance.maxVoteDays}
-        currentCoolOffHours={props.governance.coolOffHours}
-        currentMinInstructionHoldupDays={
+        initialCommunityRules={props.governance.communityTokenRules}
```
asktree, line 48: i refactored all "current" prop names to "initial" because this is more generic while also being, imo,as precise or more.

hub/components/NewWallet/useGovernanceDefaults.ts
```diff
@@ -0,0 +1,136 @@
...
+import { Rules } from '../EditWalletRules/types';
+import useTreasuryInfo from '@hooks/useTreasuryInfo';
+import { GovernanceTokenType } from '@hub/types/GovernanceTokenType';
+import { GovernanceVoteTipping } from '@hub/types/GovernanceVoteTipping';
+
+import * as RE from '@hub/types/Result';
+
+import { DISABLED_VOTER_WEIGHT } from '@tools/constants';
+
+const configs2defaults = (configs: GovernanceConfig[]) => {
```
asktree, line 18: @SebastianBor this fn takes configs from all wallets in the DAO and then computes what the strictest settings would be for the default config of a new wallet. Obviously this can result in the default configuration to be to disable all voting, but you can't actually do that so whatever. Perhaps you would like to review it.

PR #1550, opened 2023-04-17
Feature/split position

bryzettler:
Add the ability to split a position and ability to transfer an amount not just the full amount


asktree (approved):
LGTM, only loosely reviewed since it doesnt touch non-helium code.

PR #1554, opened 2023-04-18
Fix the way hybrid hub pages attempt auth

nramadas:


asktree (commented):
I'm not sure this does anything because in the hybrid components I import the JWTProvider directly, not the Root provider. and those usages of the JWTProvider havent had their props changed? Unclear.

PR #1571, opened 2023-04-23
Feature/ban unsafe imgs

abrzezinski94:


components/NewRealmWizard/components/TokenInput.tsx
```diff
@@ -82,13 +83,9 @@ export default function TokenInput({
...
-      const tokenList = await new TokenListProvider().resolve()
-      const filteredTokenList = tokenList
-        .filterByClusterSlug(
-          connection.cluster === 'mainnet' ? 'mainnet-beta' : connection.cluster
-        )
-        .getList()
-      setTokenList(filteredTokenList)
+      const tokens = await axios.get('https://token.jup.ag/strict')
+      const tokenList = tokens.data as TokenInfo[]
+      setTokenList(tokenList)
```
asktree, line 88: let's also use react-query for this because we would really prefer to cache it

hooks/queries/tokenMetadata.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const useTokenMetadata = (
+  mint: PublicKey | undefined,
+  otherConditions?: boolean
+) => {
+  const connection = useWalletStore((s) => s.connection)
+
+  const enabled = !!mint && !!otherConditions
+
+  const query = useQuery({
```
asktree, line 15: is token metadata even mutable generally? 

we can do `useQuery({cacheTime: (10 days in miliseconds), staleTime: (1hr in miliseconds)` for example, and then it will just get cached a long time and not get refreshed more than each hour

components/TreasuryAccount/AccountItem.tsx
```diff
@@ -19,63 +17,19 @@ const AccountItem = ({
...
   const [symbolFromMeta, setSymbolFromMeta] = useState<undefined | string>(
     undefined
   )
-  const connection = useWalletStore((s) => s.connection)
+  const { data } = useTokenMetadata(
+    governedAccountTokenAccount.extensions.mint?.publicKey,
+    !logo
+  )
 
   useEffect(() => {
```
asktree, line 28: just use `useMemo` here instead of a state + effect

components/TreasuryAccount/AccountsTabs.tsx
```diff
@@ -65,42 +63,20 @@ const AccountTab: FunctionComponent<AccountTabProps> = ({
...
 
-        setLogoFromMeta(data.image)
-        setSymbolFromMeta(data.symbol)
-      } catch (e) {
-        console.log(e)
-      }
-    }
-    if (!logo && assetAccount.extensions.mint?.publicKey.toBase58()) {
-      getTokenMetadata(assetAccount.extensions.mint?.publicKey.toBase58())
+  useEffect(() => {
```
asktree, line 75: same, just use `useMemo` instead of state + effect.

components/TreasuryAccount/BaseAccountHeader.tsx
```diff
@@ -16,39 +14,20 @@ const BaseAccountHeader: React.FC<{
...
-      }
-    }
     if (tokenInfo?.address) {
       setLogo(tokenInfo?.logoURI)
       setSymbol(tokenInfo?.symbol)
-    } else {
-      getTokenMetadata(mintAddress)
     }
-  })
+    if (data?.symbol) {
```
asktree, line 27: `useMemo`

components/NewRealmWizard/components/NFTCollectionSelector.tsx
```diff
@@ -16,6 +16,7 @@ function ImageWithLoader({ className, ...props }) {
       </div>
       <img
         {...props}
+        src={props.src}
```
asktree, line 19: ?

hooks/queries/tokenMetadata.ts
```diff
@@ -0,0 +1,40 @@
...
+    queryFn: async () => {
+      const mintPubkey = new PublicKey(mint!)
+      const metadataAccount = findMetadataPda(mintPubkey)
+      const accountData = await connection.current.getAccountInfo(
+        metadataAccount
+      )
+
+      const state = Metadata.deserialize(accountData!.data)
+      const jsonUri = state[0].data.uri.slice(
```
asktree, line 24: 😮 what about just using `Metadata.fromAccountAddress` and then just `result.uri` or whatever? this looks harder to maintain in future

hooks/queries/tokenMetadata.ts
```diff
@@ -0,0 +1,40 @@
...
+        0,
+        state[0].data.uri.indexOf('\x00')
+      )
+
+      const data = await (await fetch(jsonUri)).json()
+      //we dont want unsafe imgs
```
asktree, line 30: i dont follow this comment -- are imgs being filtered here?

asktree (commented):
im not sure i understand the change. it seems like the only difference is we get metadata from  jup.ag instead of the spl-token-registry thing? struggling to follow the flow of the data here

components/NewRealmWizard/components/NFTCollectionSelector.tsx
abrzezinski94: ...props are not typed and it dont recognize there is props.src

hooks/queries/tokenMetadata.ts
abrzezinski94: img is inside data. will adjust comment for people to not include imgs from metadata

abrzezinski94:
> im not sure i understand the change. it seems like the only difference is we get metadata from jup.ag instead of the spl-token-registry thing? struggling to follow the flow of the data here

i just replaced what was on front end and removed imges from metadata we can still get symbol if there is not symbol in jupiter api.

components/NewRealmWizard/components/NFTCollectionSelector.tsx
abrzezinski94: ill add htmlInputProps

hooks/queries/tokenMetadata.ts
asktree: i think it would be better if this query just returned the data in its natural state (`Metadata`) and then the components themselves were responsible for using the right source of data for the images. you could leave a comment tho. i think by default any transformation of data is bad so better to keep to a minimum. thats how i'd do it anyway.

abrzezinski94:
@asktree anyway at few places e.g treasury view this metdata rate limit rpc if you have a lot of token accounts the metadata request should be batched but it needs some refactor on wallet page.

PR #1579, opened 2023-04-25
differentiate warning states in nft wiz

asktree:
We can't verify collection status for some old collections like smb that don't use the new standard. 
previous: image

I made it look like this instead: 
image
Likewise I made the error stronger when we know it's definitely not a collection: 
image

### DIFF
--- a/components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
+++ b/components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react'
+import { useEffect, useState } from 'react'
 import { useForm, Controller } from 'react-hook-form'
 import { yupResolver } from '@hookform/resolvers/yup'
 import * as yup from 'yup'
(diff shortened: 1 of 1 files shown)


components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
```diff
@@ -463,10 +435,15 @@ export default function AddNFTCollectionForm({
...
+                  error?.message ?? collectionVerificationState === 'invalid'
+                    ? 'Error: this is not an nft collection'
+                    : ''
+                }
                 warning={
-                  unverifiedCollection
-                    ? 'Caution: we could not verify this address is a "certified" collection.'
+                  collectionVerificationState ===
+                  'is nft but no collection details'
+                    ? 'Caution: This is an nft, but has no collection details. It may be an old collection, or just a regular nft. Please double check'
```
SebastianBor, line 446: let's add `before proceeding`  to make it more explicit it's ok to go ahead if you are sure it's a correct collection address. `Please double check before proceeding`

PR #1610, opened 2023-05-01
small tweaks

bryzettler:


utils/uiTypes/VotePlugin.ts
```diff
@@ -199,6 +198,7 @@ export class VotingClient {
     }
 
     if (this.client instanceof HeliumVsrClient) {
+      console.log('WTF WHEN')
```
bryzettler, line 201:
```suggestion
```

components/VotePanel/CastVoteButtons.tsx
```diff
@@ -70,6 +94,56 @@ export const CastVoteButtons = () => {
...
+        )
+      } catch (ex) {
+        if (isNftPlugin) {
+          closeNftVotingCountingModal(
+            (client.client as unknown) as NftVoterClient,
+            proposal!,
+            wallet!.publicKey!
+          )
+        }
+        //TODO: How do we present transaction errors to users? Just the notification?
```
asktree, line 139: sendTransactionsV3 does it

components/VotePanel/CastVoteButtons.tsx
```diff
@@ -70,6 +94,56 @@ export const CastVoteButtons = () => {
...
   const isInCoolOffTime = useIsInCoolOffTime()
 
+  const handleClick = async (vote: 'yes' | 'no') => {
+    setVote(vote)
+    setSubmitting(true)
+
+    if (allowDiscussion) {
+      setShowVoteModal(true)
+    } else {
+      const confirmationCallback = async () => {
```
asktree, line 104: so is this just duplicated code from the vote modal ?

asktree: if so lets move it to a shared hook

components/chat/DiscussionPanel.tsx
```diff
@@ -1,12 +1,22 @@
+import React, { useMemo } from 'react'
 import DiscussionForm from './DiscussionForm'
 import Comment from './Comment'
 import useWalletStore from '../../stores/useWalletStore'
 
-const DiscussionPanel = () => {
+const DiscussionPanel: React.FC<any> = () => {
```
asktree, line 6: ?

components/chat/DiscussionPanel.tsx
```diff
@@ -1,12 +1,22 @@
...
   const { chatMessages, voteRecordsByVoter, proposalMint } = useWalletStore(
     (s) => s.selectedProposal
   )
 
+  const sortedMessages = useMemo(
```
asktree, line 11: wise

hooks/useRealm.tsx
```diff
@@ -244,6 +244,10 @@ export default function useRealm() {
     realmInfo,
     symbol,
     voteSymbol: realmInfo?.voteSymbol,
+    allowDiscussion:
```
asktree, line 247: `useRealm` already surfaces `realmInfo`, there is no need to create a new field. I don't like the existing way that fields are constantly re-exported like this and I'm in the process of killing it.

asktree (changes requested):
left some comments.

asktree (changes requested):
left some comments.

components/VotePanel/CastVoteButtons.tsx
```diff
@@ -58,6 +67,21 @@ const useCanVote = () => {
...
+  const wallet = useWalletOnePointOh()
+  const connection = useWalletStore((s) => s.connection)
+  const { realm, realmInfo, config, allowDiscussion } = useRealm()
+  const { refetchProposals } = useWalletStore((s) => s.actions)
+  const isNftPlugin =
+    config?.account.communityTokenConfig.voterWeightAddin &&
+    nftPluginsPks.includes(
+      config?.account.communityTokenConfig.voterWeightAddin?.toBase58()
+    )
+  const { closeNftVotingCountingModal } = useNftProposalStore.getState()
```
asktree, line 79: im confused about whats going on here, we are closing this modal but when does it get opened?

components/VotePanel/CastVoteButtons.tsx
asktree: or alternatively prop-drill it

components/VoteCommentModal.tsx
```diff
@@ -142,6 +140,14 @@ const VoteCommentModal: FunctionComponent<VoteCommentModalProps> = ({
 
   const voteString = VOTE_STRINGS[vote]
 
+  useEffect(() => {
```
asktree, line 143: I think this is probably too hacky; I really don't like it if side effects are used for business logic etcetera like this. 
What I'd rather see is the `submitVote` callback extracted to a hook and used by both this and the parent component, or simply passed as a prop to the modal by the parent.

asktree (changes requested):
see comment

components/VoteCommentModal.tsx
bryzettler: fair, ill work on extracting the logic somehow

asktree (approved):
great, thanks

PR #1678, opened 2023-06-15
replaced DL logo

DimiMili:
replaced DL logo with a smaller one (current was too huge)


asktree:
The filename is different too which seems like a mistake to me ?

PR #1723, opened 2023-07-21
use DAS api to get verified collection data in creating dao with nft-voter

hauchu1998:
Fix three things:
1. create key "totalNfts" to record the real total nfts number of a collection in NFTcollectionSelector.
2. For useDigitalAssetsByOwner, when happens error from useQuery, instead of returning it, I change it to throw it as other functions do.
3. replace getNfts old query function to the new created useOwnerVerifiedCollectionsQuery for NFT modal while creating DAO

### DIFF
--- a/components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
+++ b/components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
@@ -8,7 +8,6 @@ import { updateUserInput, validatePubkey } from '@utils/formValidation'
 import { notify } from '@utils/notifications'
 import { abbreviateAddress } from '@utils/formatting'
 
-
 import { NewButton as Button } from '@components/Button'
 import Text from '@components/Text'
 import FormHeader from '@components/NewRealmWizard/components/FormHeader'
@@ -350,10 +349,10 @@ export default function AddNFTCollectionForm({
               return enrichItemInfo(nft, nft.uri)
             })
           )
-
           verifiedCollections[collectionKey] = {
             ...collectionInfo.result,
             nfts: nftsWithInfo,
+            totalNfts: verfiedNfts[collectionKey].length,
           }
         }
       }
--- a/components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
+++ b/components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
@@ -21,27 +21,24 @@ import NFTCollectionModal from '@components/NewRealmWizard/components/NFTCollect
 import { Metaplex } from '@metaplex-foundation/js'
 import { Connection, PublicKey } from '@solana/web3.js'
 import useWalletOnePointOh from '@hooks/useWalletOnePointOh'
-import asFindable from '@utils/queries/asFindable'
 import useLegacyConnectionContext from '@hooks/useLegacyConnectionContext'
 
-function filterAndMapVerifiedCollections(nfts) {
+export function filterAndMapVerifiedCollections(nfts) {
   return nfts
-    .filter((nft) => {
-      if (nft.data?.collection) {
-        return nft.data?.collection?.verified
+    ?.filter((nft) => {
+      if (
+        nft.grouping &&
+        nft.grouping.find((x) => x.group_key === 'collection')
+      ) {
+        return true
       } else {
-        return nft.collection?.verified
-      }
-    })
-    .map((nft) => {
-      if (nft.data?.collection) {
-        return nft.data
-      } else {
-        return nft
+        return false
       }
     })
     .reduce((prev, curr) => {
-      const collectionKey = curr.collection?.key
+      const collectionKey = curr.grouping.find(
+        (x) => x.group_key === 'collection'
+      )?.group_value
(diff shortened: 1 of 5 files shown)


hooks/queries/digitalAssets.ts
```diff
@@ -185,8 +185,11 @@ export const useDigitalAssetsByOwner = (owner: undefined | PublicKey) => {
   return useQuery({
     enabled,
     queryKey: owner && digitalAssetsQueryKeys.byOwner(network, owner),
-    queryFn: async () =>
-      enabled ? dasByOwnerQueryFn(network, owner) : new Error(),
+    queryFn: async () => {
```
asktree, line 188: great

hooks/queries/nft.ts
```diff
@@ -57,3 +61,40 @@ export const fetchNFTbyMint = (connection: Connection, pubkey: PublicKey) => {
...
+  const verifiedNfts = filterAndMapVerifiedCollections(ownedNfts)
+
+  const enabled = owner !== undefined && ownedNfts !== undefined
+  const query = useQuery({
+    queryKey: enabled ? nftQueryKeys.byMint(network, owner) : undefined,
+    queryFn: async () => {
+      if (!enabled) throw new Error()
+
+      const verifiedCollections = {}
+      for (const collectionKey in verifiedNfts) {
```
asktree, line 78: it would be better to do this in parallel using Promise.all

hooks/queries/nft.ts
```diff
@@ -57,3 +61,40 @@ export const fetchNFTbyMint = (connection: Connection, pubkey: PublicKey) => {
...
 }
+
+export const useOwnerVerifiedCollectionsQuery = (owner: PublicKey) => {
+  const { connection } = useConnection()
+  const { data: ownedNfts } = useDigitalAssetsByOwner(owner)
+  const network = getNetworkFromEndpoint(connection.rpcEndpoint)
+  const verifiedNfts = filterAndMapVerifiedCollections(ownedNfts)
+
+  const enabled = owner !== undefined && ownedNfts !== undefined
+  const query = useQuery({
```
asktree, line 72: I'm not sure it really makes sense for this to use useQuery because there's nothing to cache at the owner level. 

Here's what we want cached:
- what NFTs `owner` owns (thats cached by useDigitalAssetsByOwner)
- the collection nft data (that's cacehd by `fetchNFTbyMint`)

So there is nothing else to cache or key. Instead, you can just use `useAsync` for this and no key.

asktree: And you could just put this in regular hooks folder, or even just in the one file it gets used in (I assume its just one? the wizard?)

components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
```diff
@@ -519,7 +472,7 @@ export default function AddNFTCollectionForm({
                       .map((nft, index) => {
                         return (
                           <img
-                            key={nft.name}
+                            key={nft.name + index}
```
asktree, line 475: nit: surely should key on nft mint ?

components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
```diff
@@ -22,27 +21,24 @@ import NFTCollectionModal from '@components/NewRealmWizard/components/NFTCollect
 import { Metaplex } from '@metaplex-foundation/js'
 import { Connection, PublicKey } from '@solana/web3.js'
 import useWalletOnePointOh from '@hooks/useWalletOnePointOh'
-import asFindable from '@utils/queries/asFindable'
 import useLegacyConnectionContext from '@hooks/useLegacyConnectionContext'
 
-function filterAndMapVerifiedCollections(nfts) {
+export function filterAndMapVerifiedCollections(nfts) {
```
asktree, line 26: this only gets used one time, so it should just be colocated with the one time it gets used (wherever you end up putting that). this matters for avoiding circular deps, which the build system usually can cope with but sometimes cant.

components/NewRealmWizard/components/steps/AddNFTCollectionForm.tsx
```diff
@@ -323,59 +319,14 @@ export default function AddNFTCollectionForm({
...
     try {
       setWalletConnecting(true)
-
       if (!connected) {
         if (wallet) await wallet.connect()
       }
       if (!wallet?.publicKey) {
         throw new Error('No valid wallet connected')
       }
-      const metaplex = new Metaplex(connection.current)
```
asktree, line 333: perfect

hooks/queries/nft.ts
hauchu1998: Changed to Promise.all

hooks/queries/nft.ts
hauchu1998: modify useQuesry to useAync instead.

asktree (approved):
LGTM!

hauchu1998:
use DAS api to get verified collection data in creating dao with nft-voter

hauchu1998:
merge

PR #1737, opened 2023-08-01
stop storing TOR pubkey in url for account page

asktree:
previously you viewed your account at 

blablah/account/[tokenOwnerRecordPk]

so if you switched wallets, this didn't update. the reason this was used was so that you could easily view other user's accounts via link (only VSR does this).

now normal usage uses blahblah/account/me, and you can do blahblah/account/[walletPk] in vsr to view other accounts.


asktree:
tested on vanilla dao, mngo, and tested the [walletPk] page on MNGO

PR #1747, opened 2023-08-04
Prop. Explorer fix

kirksgithub:
Fixed proposal explore router to correctly navigate back to a proposal rather than using the browser (back) routing which breaks in direct links


pages/dao/[symbol]/proposal/[pk]/explore.tsx
```diff
@@ -38,7 +38,7 @@ export default function Explore() {
           'transition-all',
           'hover:text-fgd-3'
         )}
-        onClick={router.back}
+        onCLick={router.asPath.replace(/\/explore$/, '')}
```
Comment (asktree) [inline pages/dao/[symbol]/proposal/[pk]/explore.tsx:41]:
```suggestion
        onClick={router.asPath.replace(/\/explore$/, '')}
```

PR #1771, opened 2023-08-15
added logo for wikicious DAO

hetwik:


hetwik (commented):
Added wikicious dao info

PR #1777, opened 2023-08-16
fix createLUTProposal tx size

hauchu1998:
While extending lookup table, the max address to add at once]e is ~20 addresses. 
reference from [here](https://docs.solana.com/developing/lookup-tables#:~:text=NOTE%3A%20Due%20to,transaction%27s%20memory%20limits).

Since sending cNFT can easily exceed that amount due to the proof and other instruction accounts, splitting `extendInstruction` into multiple transactions is necessary. Moreover, maybe `createLookupTable` and `extendLookupTable` should also separate into different transactions or the tx size will be larger than the restriction?

### DIFF
--- a/actions/createLUTproposal.ts
+++ b/actions/createLUTproposal.ts
@@ -123,8 +123,7 @@ export const createLUTProposal = async (
     .map((x) => x.chunkBy!)
 
   const lowestChunkBy = chunkBys.length ? Math.min(...chunkBys) : 2
-  console.log(instructionsData)
-  console.log('lowestChunkBy', lowestChunkBy)
+
   for (const [index, instruction] of instructionsData
     .filter((x) => x.data)
     .entries()) {
--- a/actions/createLUTproposal.ts
+++ b/actions/createLUTproposal.ts
@@ -123,7 +123,8 @@ export const createLUTProposal = async (
     .map((x) => x.chunkBy!)
 
   const lowestChunkBy = chunkBys.length ? Math.min(...chunkBys) : 2
-
+  console.log(instructionsData)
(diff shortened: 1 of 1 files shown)


actions/createLUTproposal.ts
```diff
@@ -246,12 +257,15 @@ export const createLUTProposal = async (
...
     transactionInstructions: [
       {
-        instructionsSet: [
-          { transactionInstruction: lookupTableInst },
-          { transactionInstruction: extendInstruction },
-        ],
+        instructionsSet: [{ transactionInstruction: lookupTableInst }],
         sequenceType: SequenceType.Sequential,
       },
+      ...extendInstructions.map((x) => {
```
hauchu1998, line 263: should this be 
```
...extendInstructions.map((x, idx) => {
        return {
          instructionsSet: [{ transactionInstruction: x }],
          sequenceType: idx == 0? SequenceType.Sequential : SequenceType.Parallel,
        }
      }),
```
to prevent parallel tx won't wait till sequential tx is completed?
However, when I'm testing sendNFT, the PR code works.

actions/createLUTproposal.ts
```diff
@@ -246,12 +256,15 @@ export const createLUTProposal = async (
...
+        return {
+          instructionsSet: [{ transactionInstruction: x }],
+          sequenceType: SequenceType.Parallel,
```
asktree, line 265:
```suggestion
          sequenceType: SequenceType.Sequential,
```

actions/createLUTproposal.ts
asktree: technically this could be in parallel but, because of reasons, i'd rather it be sequential.

PR #1782, opened 2023-08-19
refactor voting power

asktree:
Currently each plugin has a zustand store that keeps track of voting power. This is already annoying, but it also means that the current data pipeline only supports getting voting power of a single user. This doesn't gel with the ongoing rework of the delegation UX, so it needs to change.

I'm getting rid of the zustand stores and replacing them with queries that get voting power. I want to do this incrementally, so for now I am still getting the data from them, but this PR sets things up such that each store can be removed one at a time.

This PR also removes voting power from the nft-voter zustand store, and uses react-query queries instead.


asktree:
I tested using Helium & Mango daos to make sure I didn't break other plugins.

asktree:
I also tested an NFT dao

PR #1785, opened 2023-08-21
Dual finance changes, validation and preview of strikes on Staking Option

brittcyr:


abrzezinski94:
got some ts error that prevent deployment 
Zrzut ekranu 2023-08-23 o 10 33 13

brittcyr:
> got some ts error that prevent deployment Zrzut ekranu 2023-08-23 o 10 33 13

Screen Shot 2023-08-23 at 9 04 03 AM
I thought that was fixed by the most recent commit, but I will add more explicit checks

PR #1791, opened 2023-08-23
fix versionedTransaction bug affecting helium via squads

asktree:


asktree:
closing because the problem is just lack of support from squads

PR #1813, opened 2023-09-08
add ResonanceDAO metadata

ilovespectra:


ilovespectra:
i was unable to create token metadata with this dao, so i created a new ResonanceDAO and have succeeded 2 proposals to add the community and council token metadatas. I'm being told i have insufficient funds for either despite having several SOL in both council wallets. i may be in need of your help if you could please take a look. thank you!

PR #1822, opened 2023-09-13
VSR: unlock deposit integration

tomjohn1028:
* Updates the VSR IDL to 0.2.5, which is built from [this commit](https://github.com/mithraiclabs/voter-stake-registry/commit/5472db4e022caad01b5e68e3cf80820a9e61265b)
*Add support for the `UnlockDeposit` instruction

### DIFF
--- a/VoteStakeRegistry/actions/getUnlockDepositInstruction.ts
+++ b/VoteStakeRegistry/actions/getUnlockDepositInstruction.ts
@@ -1,0 +1,45 @@
+import { PublicKey } from '@solana/web3.js'
+
+import { getRegistrarPDA, getVoterPDA } from 'VoteStakeRegistry/sdk/accounts'
+import { VsrClient } from 'VoteStakeRegistry/sdk/client'
+
+export const getUnlockDepositInstruction = async ({
+  grantAuthorityPk,
+  voterAuthorityPk,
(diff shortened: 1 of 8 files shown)


asktree:
Is there some way we can get the VSR version / fork using on-chain data? this stuff shouldnt show up if the version of VSR doesn't support it.

To be clear, you are running a VSR fork right? this isn't some new feature in mainline VSR?

tomjohn1028:
> Is there some way we can get the VSR version / fork using on-chain data? this stuff shouldnt show up if the version of VSR doesn't support it.

The most fool proof way would be a hard coded list of program ID's that feature gate this instruction.

> To be clear, you are running a VSR fork right? this isn't some new feature in mainline VSR?

Correct. Discussion on whether these features getting added to main VSR is TBD. So for now it's best to assume it will not make it into the main VSR repo.

SebastianBor:
> Correct. Discussion on whether these features getting added to main VSR is TBD. So for now it's best to assume it will not make it into the main VSR repo.

Do you have more details what the new features are about please?

tomjohn1028:
@SebastianBor added more details to the comment in this PR https://github.com/mithraiclabs/voter-stake-registry/pull/1#issue-1865895596

PR #1856, opened 2023-10-05
EditWalletRules: Integer quorum percent

asktree:
Quorum percent can only be an integer between 1 and 100%, so this PR updates the wallet rules form to reflect this.


asktree:
actually 1% may be the minimum

PR #1875, opened 2023-10-16
Pre-populating the Deposit Max Amount

kirksgithub:
The Deposit button automatically allowed the Max amount to be deposited, nevertheless this changed causing issues for our DAO members, in specific we identified the following:
- Deposit amount should be defaulted to the Max amount
- Deposit amount when typing is no longer available, so for larger numbers the total available, the user can enter an amount which is not their actual max amount.

To improve this slow and confusing experience, the popup should be easier to navigate as specified above.

We also recommend that in the future this button is pushed as an advanced button either in the "My Governance View" or as a dropdown button as an additional option

The fixes here will be to improve the UI/UX and allow a default Max to be set along with any helper buttons to ease setting the Max deposit amount

### DIFF
--- a/components/DepositTokensButton.tsx
+++ b/components/DepositTokensButton.tsx
@@ -89,7 +89,7 @@ export const DepositTokensButton = ({
                 await deposit(nativeAmount)
                 setOpenModal(false)
               }}
-              disabled={(humanReadableMax && parseInt(amount) > humanReadableMax)}
+              disabled={humanReadableMax !== undefined && parseInt(amount) > humanReadableMax}
             >
               Confirm
             </Button>
--- a/components/DepositTokensButton.tsx
+++ b/components/DepositTokensButton.tsx
@@ -70,7 +70,7 @@ export const DepositTokensButton = ({
               Amount to deposit
               <span>
                 &nbsp;-&nbsp;<a href="#" onClick={() => { setAmount(humanReadableMax ? humanReadableMax.toString() : '') }}>Max</a>
-              </span>)
+              </span>
             </label>
             <Input
               placeholder={humanReadableMax?.toString() + ' (max)'}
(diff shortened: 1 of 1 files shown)


kirksgithub:
@asktree this update fixes UI/UX issues with depositing tokens to a governance, although this is a button should have been pushed as in an advanced menu in the "My Governance Power" and not exposed for regular users who would like to quickly deposit and participate in Governance:

- Added Max helper button amount
- Set default input to have the max amount (as it is typically intended by the user to deposit the entire amount to governance)
- Label prop was moved outside the input field to support a Max button

This should improve the experience significantly so users are not guessing the amount they have as a balance while in this prompt

components/DepositTokensButton.tsx
```diff
@@ -76,6 +88,7 @@ export const DepositTokensButton = ({
                 await deposit(nativeAmount)
                 setOpenModal(false)
               }}
+              disabled={humanReadableMax !== undefined && (parseInt(amount) > humanReadableMax || parseInt(amount) > 0)}
```
asktree, line 91: this seems wrong? if `amount > 0` the button is disabled?

asktree:
The intended UX is that the max is a placeholder, and hitting confirm without any value entered would use it. It's just an oversight in the code that this isn't the case. However, it seems like this intended UX wasn't very obvious. I'm going to fix this bug in another PR quickly. I think the ideal UI would be a 1% to 100% slider, with a text entry box.

kirksgithub:
Default should be set to Max, this PR resolves the issue immediately until you push your changes

components/DepositTokensButton.tsx
```diff
@@ -30,12 +30,11 @@ export const DepositTokensButton = ({
   const depositTooltipContent = !connected
     ? 'Connect your wallet to deposit'
     : !hasTokensInWallet
-    ? "You don't have any governance tokens in your wallet to deposit."
-    : undefined
+      ? "You don't have any governance tokens in your wallet to deposit."
```
asktree, line 33: it seems like you're using a different code formatter, any idea what's going on? FWIW I prefer this way of formatting ternaries

kirksgithub: using code spaces

asktree: as in, using the codespace associated with the repo? interesting. how did you find that btw? I'm new to codespaces and just set it up for personal use.

PR #1880, opened 2023-10-16
Update DepositTokensButton.tsx

kirksgithub:
Fixed Disabled state on button

### DIFF
--- a/components/DepositTokensButton.tsx
+++ b/components/DepositTokensButton.tsx
@@ -5,7 +5,7 @@ import useUserGovTokenAccountQuery from '@hooks/useUserGovTokenAccount'
 import { useDepositCallback } from './GovernancePower/Vanilla/useDepositCallback'
 import useWalletOnePointOh from '@hooks/useWalletOnePointOh'
 import Modal from './Modal'
-import { useState } from 'react'
+import { useState, useEffect } from 'react'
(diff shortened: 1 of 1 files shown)


kirksgithub:
The Deposit button automatically allowed the Max amount to be deposited, nevertheless this changed causing issues for our DAO members, in specific we identified the following:

Deposit amount should be defaulted to the Max amount
Deposit amount when typing is no longer available, so for larger numbers the total available, the user can enter an amount which is not their actual max amount.
To improve this slow and confusing experience, the popup should be easier to navigate as specified above.

We also recommend that in the future this button is pushed as an advanced button either in the "My Governance View" or as a dropdown button as an additional option

The fixes here will be to improve the UI/UX and allow a default Max to be set along with any helper buttons to ease setting the Max deposit amount

kirksgithub:
This update fixes UI/UX issues with depositing tokens to a governance, although this button should have been pushed as an advanced menu option/deposit option in the "My Governance Power" and not exposed for regular users who would like to quickly deposit and participate in Governance:

Added Max helper button amount
Set default input to have the max amount (as it is typically intended by the user to deposit the entire amount to governance)
Disable confirm button if amount <= 0 and amount > max
Label prop was moved outside the input field to support a Max button
This should improve the experience significantly so users are not guessing the amount they have as a balance while in this prompt

(sorry for the second commit)

PR #1918, opened 2023-11-12
Bring back PYTH DAO

guibescos:
We're launching Pyth DAO soon (on Monday), we are using our custom voter weight plugin. 
Pyth DAO got removed here https://github.com/solana-labs/governance-ui/pull/1752 , this PR is mostly reverting those changes and cleaning up some weird/dirty things we had done back in the day.

@asktree what would it take to merge this?


utils/uiTypes/VotePlugin.ts
```diff
@@ -148,21 +151,26 @@ export class VotingClient {
       this.clientType = VotingClientType.GatewayClient
       this.noClient = false
     }
+    if (this.client instanceof PythClient) {
```
asktree, line 154: instead of adding onto a monolithic class, how about making a new class that implements a common interface

hooks/queries/governancePower.ts
```diff
@@ -102,6 +106,20 @@ export const getNftGovpower = async (
   return power
 }
 
+export const getPythGovPower = async (connection: Connection, user : PublicKey | undefined) : Promise<BN> => {
+  if (!user) return new BN(0)
+
+  const pythClient = await PythClient.connect(connection, new NodeWallet(new Keypair()))
```
guibescos, line 112: This client is readonly and I don't know any more elegant way of creating a "random wallet"

utils/uiTypes/VotePlugin.ts
```diff
@@ -148,21 +151,26 @@ export class VotingClient {
...
-    createNftActionTicketIxs?: TransactionInstruction[]
+    createNftActionTicketIxs?: TransactionInstruction[],
+    voterWeightTarget?: PublicKey
   ): Promise<ProgramAddresses | undefined> => {
     if (this.noClient) return
-
     const realm = this.realm!
     const torAccount = await fetchTokenOwnerRecordByPubkey(
-      this.client!.program.provider.connection,
+      this.client!.stakeConnection.program.provider.connection, // FIX ME
```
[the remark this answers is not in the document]
guibescos: It's fixed!

utils/uiTypes/VotePlugin.ts
guibescos: I really don't understand what this means, do you care to explain or give an example of someone that integrated this way?

utils/uiTypes/VotePlugin.ts
```diff
@@ -148,12 +151,17 @@ export class VotingClient {
...
+      this.noClient = false
+    }
   }
   withUpdateVoterWeightRecord = async (
     instructions: TransactionInstruction[],
     tokenOwnerRecord: PublicKey,
     type: UpdateVoterWeightRecordTypes,
-    createNftActionTicketIxs?: TransactionInstruction[]
+    createNftActionTicketIxs?: TransactionInstruction[],
+    voterWeightTarget?: PublicKey
```
asktree, line 164: It's problematic that this parameter is not clearly pyth specific.
```suggestion
    pythVoterWeightTarget?: PublicKey
```

hooks/queries/governancePower.ts
asktree: Yeah I mean it seems like a wallet should not actually be required by your SDK. But this seems sufficiently elegant as a way to pass a dummy wallet. You can also just like pass `{} as NodeWallet`.

hooks/useVotingPlugins.ts
```diff
@@ -137,6 +141,9 @@ export function useVotingPlugins() {
         if (HELIUM_VSR_PLUGINS_PKS.includes(currentPluginPk.toBase58())) {
           handleSetHeliumVsrClient(wallet, connection, currentPluginPk)
         }
+        if (PYTH_PLUGIN_PK.includes(currentPluginPk.toBase58())) {
```
asktree, line 144: there is really no reason that all the plugins should be in the same `useEffect`, and many reasons they should not be. Please just make one `useEffect` for initializing your plugin. And please give that `useEffect` exhaustive dependencies.

0xfirefist: addressed

utils/uiTypes/VotePlugin.ts
0xfirefist: renamed based on the suggestion

components/TokenBalance/TokenDeposit.tsx
```diff
@@ -325,7 +315,7 @@ export const TokenDeposit = ({
           </div>
 
           <div className="flex flex-col mt-6 space-y-4 sm:flex-row sm:space-x-4 sm:space-y-0">
-            {hasTokensInWallet || inAccountDetails ? (
+            {/* {hasTokensInWallet || inAccountDetails ? (
```
asktree, line 318: explain why? this may make sense but just checking

asktree: yeah I just looked over this and it will surely break things?

asktree: I know the component hierarchy here is annoying, but this component is used in many places, and hiding the deposit button in all of them isnt going to be right.

montana77783 (approved):

montana77783:
https://snowtrace.io/token/0x9702230A8Ea53601f5cD2dc00fDBc13d4dF4A8c7/contract/code?chainid=43114#abi

PR #1930, opened 2023-11-19
fix typos

omahs:
fix typos


asktree:
I believe this file is only used by realms hubs and therefore not in use at all any longer.

PR #2006, opened 2023-12-24
Add extra validation

shiralwz:
1. Trim the address before passing it into the validatePublicKey function.
2. Check if the val parameter is a string and a valid public key before further processing
3. check whether solAccount is defined before accessing its lamports property.

### DIFF
--- a/utils/validations.tsx
+++ b/utils/validations.tsx
@@ -99,6 +99,10 @@ const validateDestinationAccAddress = async (
   val: any,
   governedAccount?: PublicKey
 ) => {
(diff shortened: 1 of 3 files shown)


utils/validations.tsx
```diff
@@ -99,6 +99,10 @@ const validateDestinationAccAddress = async (
   val: any,
   governedAccount?: PublicKey
 ) => {
+  if (typeof val !== 'string' || !validatePubkey(val)) {
```
asktree, line 102: why not just change the param type of `val` to string, and let TS handle this?

asktree (changes requested):
see comment

PR #2026, opened 2024-01-03
[pyth] Add proposal votes scaling for Pyth votes

guibescos:
In the Pyth Realm only staked tokens can vote and the voting power of each staked token is inversely proportional to how many tokens are currently staked (so that at any time the sum of the power of all staked tokens is equal to the total voting power of 10 Billion).
In the UI, voting power gets displayed, leading to a lot of confusion in the DAO.
It would be better to do a change of unit and display things as the number of staked tokens.
This is what this PR is attempting.


hooks/PythNetwork/useScalingFactor.ts
```diff
@@ -0,0 +1,33 @@
+import NodeWallet from "@coral-xyz/anchor/dist/cjs/nodewallet";
+import { determineVotingPowerType } from "@hooks/queries/governancePower";
+import { useRealmQuery } from "@hooks/queries/realm";
+import { PythClient } from "@pythnetwork/staking";
+import { useConnection } from "@solana/wallet-adapter-react";
+import { useAsync } from "react-async-hook";
+
+
+export default function useScalingFactor() : number | undefined {
+    const realm = useRealmQuery().data?.result
```
0xfirefist, line 10: If we are using realm for only it's publicKey, can we use `useSelectedRealmPubkey`? I am assuming this returns the same publicKey.

hooks/PythNetwork/useScalingFactor.ts
```diff
@@ -0,0 +1,33 @@
...
+    const {connection} = useConnection()
+    const { result: plugin } = useAsync(
```
0xfirefist, line 12: Yes, I agree this would calculate scaling factor for each proposal. I think we should use useQuery instead. That will help us calculate scaling factor only once and then reuse it everywhere.

guibescos: Done

package.json
```diff
@@ -71,7 +71,7 @@
     "@project-serum/serum": "0.13.65",
     "@project-serum/sol-wallet-adapter": "0.2.6",
     "@pythnetwork/client": "2.17.0",
-    "@pythnetwork/staking": "^2.0.0",
+    "@pythnetwork/staking": "^2.1.0",
```
guibescos, line 74: Bump pythnetwork/staking package

hooks/PythNetwork/useScalingFactor.ts
guibescos: Done

hooks/PythNetwork/useScalingFactor.ts
```diff
@@ -0,0 +1,33 @@
...
+    )
+
+    const { data: scalingFactor } = useQuery(["pyth-scaling-factor"],
+        async (): Promise<number> => {
+            const pythClient = await PythClient.connect(connection, {} as NodeWallet)
+            return pythClient.getScalingFactor()
+        }, { enabled: plugin == "pyth" })
+
+    if (plugin == "pyth") {
+        return scalingFactor || 1
```
asktree, line 28: I think it's better to avoid returning fake data and just return undefined while loading. Per-component you can decide what to do with this.

hooks/useProposalVotes.tsx
```diff
@@ -100,12 +103,12 @@ export default function useProposalVotes(proposal?: Proposal) {
     voteThresholdPct,
     yesVotePct,
     yesVoteProgress,
-    yesVoteCount,
-    noVoteCount,
+    yesVoteCount : Math.floor(yesVoteCount * scalingFactor),
```
asktree, line 106: if you are going to add custom logic to generic calculations like this, it should at least be VERY explicit at each stage that A) scalingFactor is a Pyth thing and B) it's going to == 1 if Pyth isn't being used. Even more explicit would be to make it simply undfined if pyth isn't used, and then in these lines write `(pythScalingFactor || 1)`.

hooks/PythNetwork/useScalingFactor.ts
guibescos: Solved

hooks/useProposalVotes.tsx
guibescos: Solved

PR #2041, opened 2024-01-08
[pyth] Load max voter weight 

guibescos:
This was making finalizing proposals impossible, it is static address so it makes sense to load it here.


asktree (approved):
Sure. FYI, soon I will be migrating all `maxVoterWeight` out of zustand, but you dont really need to do anything for that.

PR #2053, opened 2024-01-11
[WIP] Feature/qv 7  quadratic support

dankelleher:
Since QV needs to be paired up with some other plugin that protects against sybil attacks, it will typically have a predecessor plugin (e.g. the Civic Gateway plugin)

In milestone 3 of the QV work, there will be a way to configure this via the UI.

In the meantime, we use the heuristic that the existing plugin should be used as the predecessor.

If this is not desired, for any reason, then the workaround is to move back to "default" first.

We also only enable default configuration of the quadratic plugin, resulting in a vote weight that is the square root of the token balance.

image


GatewayPlugin/sdk/api.ts
```diff
@@ -1,16 +1,78 @@
...
 
+// Get the registrar account for a given realm
 export const tryGetGatewayRegistrar = async (
   registrarPk: PublicKey,
-  client: GatewayClient
+  gatewayClient: GatewayClient
 ) => {
   try {
-    const existingRegistrar = await client.program.account.registrar.fetch(
+    return await gatewayClient.program.account.registrar.fetch(
```
asktree, line 12: is this cached?

asktree:
LGTM except I would like to see caching of any queries, using react-query

GatewayPlugin/sdk/api.ts
chriteixeira: I will update the PR to cache it.

QuadraticPlugin/sdk/accounts.tsx
```diff
@@ -0,0 +1,154 @@
...
+    client: QuadraticClient,
+    realm: ProgramAccount<Realm>
+): Promise<Coefficients> => {
+  // Get the registrar for the realm
+  const { registrar } = await getRegistrarPDA(
+      realm.pubkey,
+      realm.account.communityMint,
+      client.program.programId
+  )
+  const registrarObject = await client.program.account.registrar.fetch(
```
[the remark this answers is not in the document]
chriteixeira: I will update the PR to cache it.

PR #2098, opened 2024-02-01
[Pyth] Display voting power properly

guibescos:
This PR introduces PythVotingPower so Pyth users can see their voting power (powered by a vote plugin) when in the Pyth DAO.
For the Pyth DAO, it also disables the deposit button on realms (users have to deposit on staking.pyth.network). The withdraw button remains available for people that accidentally deposited in the past.


components/TokenBalance/TokenDeposit.tsx
```diff
@@ -101,29 +103,45 @@ export const TokenDeposit = ({
...
+      <div
+        className={`my-4 opacity-70 text-xs  ${(plugin === 'pyth') ? 'block' : 'hidden'
+          }`}
+      >
+        If you previously deposited tokens on this page, use the button below to withdraw them immediately. Those tokens have no voting power.
       </div>
 
       <div className="flex flex-col mt-6 space-y-4 sm:flex-row sm:space-x-4 sm:space-y-0">
-        {hasTokensInWallet || inAccountDetails ? (
+        {(hasTokensInWallet || inAccountDetails) && plugin !== "pyth" ? (
```
guibescos, line 144: Disable Deposit token for Pyth, since they need to deposit on our staking website.

components/TokenBalance/TokenDeposit.tsx
```diff
@@ -16,6 +16,8 @@ import { useRealmConfigQuery } from '@hooks/queries/realmConfig'
 import VanillaVotingPower from '@components/GovernancePower/Vanilla/VanillaVotingPower'
 import { DepositTokensButton } from '@components/DepositTokensButton'
 import VanillaWithdrawTokensButton from './VanillaWithdrawTokensButton'
+import { findPluginName } from '@constants/plugins'
+import PythVotingPower, { PYTH_INSTRUCTIONS } from '@components/ProposalVotingPower/PythVotingPower'
 
 /** deposit + withdraw for vanilla govtokens, used only in account view. plugin views still use this for council. */
```
asktree, line 22: I think you should leave this component as vanilla and make a Pyth version instead, and use that. (I didn't see where you actually import this component anyway, but the hierarchy is very confusing here and that's not your fault.)

PR #2119, opened 2024-02-08
Dual lockup option update to config v2

brittcyr:


brittcyr:
Lock file needs to be updated

brittcyr:
> Lock file needs to be updated

It never works for me when I do that locally

brittcyr:
> > Lock file needs to be updated
> 
> It never works for me when I do that locally

fixed

PR #2169, opened 2024-03-04
Update code to use ENV vars for Jupiter swap api #77

zer0cache:
Allows for the Jupiter V6 Swap API url to be configured via an environment variable while defaulting to the original hardcoded value and document the use of it. This should be helpful for cases where customers run into rate limits with the public API or are looking for lower latency with self/paid hosted installations.

Changes:
- Makes the Jupiter Swap V6 API base URL configurable
- Changes default URL to public.jupiterapi.com with higher rate limits. (see [jupiterapi.com](https://www.jupiterapi.com))
- Updates the README


.env.sample
```diff
@@ -5,6 +5,7 @@ DEVNET_RPC=https://mango.devnet.rpcpool.com
 
 DEFAULT_GOVERNANCE_PROGRAM_ID=GTesTBiEWE32WHXXE2S4XbZvA5CrEc4xs6ZgRe895dP
 
+NEXT_PUBLIC_JUPTER_SWAP_API_ENDPOINT=https://public.jupiterapi.com
```
abrzezinski94, line 8: why not default to quote api ?

PR #2185, opened 2024-03-14
Refactor/drop heliumvsr support

bryzettler:
Dropping all refrences of HeliumVsr and showing a Helium specific overlay when on Helium dao's and subDao's.

image


defihook (commented):
approved

PR #2200, opened 2024-03-22
Add  MailDAO

solmail:
SolMail is a groundbreaking communication protocol that enables users to send and receive mail using only a wallet address.


defihook (commented):
approved

PR #2263, opened 2024-04-24
fix: restore target in create proposal

guibescos:
Pyth's withUpdateVoterWeightRecord needs to receive the address of the governance with `createProposal`, this is because it checks the `max_voting_time` field int he governance account.


asktree:
makes sense

PR #2284, opened 2024-05-04
add RZ METAVERSITY DAO

rz1989s:
add RZ METAVERSITY DAO:

- add RZ Folder Assets under realms
- modified realms/mainnet-beta.json


defihook (commented):
great work

PR #2347, opened 2024-06-02
Correct the `exists` function to properly check both null and undefined

orion-047:
This PR fixes a bug in the `exists` function within the `GovernedAccountSelect` component, which previously returned `true` when the `item` is `undefined`.


defihook (commented):
great work

PR #2368, opened 2024-06-19
Panos/leaderboard

ppsimatikas:
image

This PR is adding a Leaderboard for all the member of the DAO with a voting power.
This is a cool functionality that Dean's list DAO needed specifically.

I figured I could add it in the Realms repo so other DAOs can benefit from it.


defihook (commented):
great work

PR #2399, opened 2024-08-07
Fix the start unlock error

ppsimatikas:
When the user tries to unlock their tokens.

image

They get this error:

image

There was a force check `!` of an undefined value, I changed it to a soft check `?`.
This will just wait for the component to load the `realmTokenAccount`


defihook (commented):
great work

